PR #47, opened 2023-09-18
Fix bug in navigation task

jysdoran:
When the goal probability is one the whole array should be true instead of false to allow for rewards

### DIFF
--- a/navix/tasks.py
+++ b/navix/tasks.py
@@ -50,13 +50,15 @@ def navigation(prev_state: State, action: Array, state: State) -> Array:
     reached = jax.vmap(jnp.array_equal, in_axes=(None, 0))(
         player.position, goals.position
     )
-    any_reached = jnp.sum(reached)
 
-    draw = jax.random.uniform(state.key, ())
-    reward = any_reached * jnp.less_equal(draw, goals.probability)
-    reward = jnp.asarray(reward, jnp.float32).squeeze()
+    # exoxgenous reward noise
+    draws = jax.random.uniform(state.key, (len(goals.probability),))
+    # if there is a reward and the player reached the goal, then reward
+    reward = reached * jnp.less_equal(draws, goals.probability)
+    # if two goals are at the same position, we sum them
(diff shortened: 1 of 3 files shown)


epignatelli:
Thank you, @jysdoran, for spotting this!

navix/tasks.py
```diff
@@ -53,7 +53,7 @@ def navigation(prev_state: State, action: Array, state: State) -> Array:
     any_reached = jnp.sum(reached)
 
     draw = jax.random.uniform(state.key, ())
```
epignatelli, line 55: Maybe we can also switch to each individual goal depending on its own marginal distribution?
```suggestion
    draws= jax.random.uniform(state.key, (len(goals.probability),))
```

navix/tasks.py
```diff
@@ -53,7 +53,7 @@ def navigation(prev_state: State, action: Array, state: State) -> Array:
     any_reached = jnp.sum(reached)
```
epignatelli, line 53:
```suggestion
    
```

navix/tasks.py
```diff
@@ -53,7 +53,7 @@ def navigation(prev_state: State, action: Array, state: State) -> Array:
...
 
     draw = jax.random.uniform(state.key, ())
-    reward = any_reached * jnp.greater_equal(draw, goals.probability)
+    reward = any_reached * jnp.less_equal(draw, goals.probability)
     reward = jnp.asarray(reward, jnp.float32).squeeze()
```
epignatelli, line 57:
```suggestion
    reward = jnp.asarray(jnp.sum(reward), jnp.float32).squeeze()
```

epignatelli (changes requested):
Thanks, @jysdoran, for spotting the issue!
I found another bug while reviewing this -- added some comments.

I pushed a uniitest to make sure this remains fixed in the future.

epignatelli (dismissed):

epignatelli:
Thanks @jysdoran for the PR again, I just pushed the other fixes, add the unitest and bumped the version to `0.3.11`.

The fix should come online in a few minutes.

PR #79, opened 2024-07-03
purejaxrl example

luchris429:
Example files for integrating cleanrl / purejaxrl-style PPO implementation.

Should be identical to the [purejaxrl implementation](https://github.com/luchris429/purejaxrl/blob/main/purejaxrl/ppo_minigrid.py).


epignatelli:
Thanks @luchris429 for pushing this!

I'll update the version to the latest available.
Other than that LGTM!

epignatelli (changes requested):
Thanks @luchris429 for pushing this!

Could you update the version to the latest available (`0.6.15` now, or whatever is the latest at the time you'll read this)?

Other than that LGTM!

epignatelli (approved):
LGTM! Thank yo uvery much, @luchris429 for pushing this!

PR #117, opened 2026-08-14
feat: disable_autoreset flag + FourRooms size variants

epignatelli:
## Summary
Ports two clean, self-contained additions from [p-doom/navix](https://github.com/p-doom/navix) (found while scouting forks for portable work), cherry-picked with original authorship preserved:

- `disable_autoreset: bool` field on `Environment` (defaults `False`, fully backward-compatible). When set, `step()` skips the auto-reset-on-done behavior, so a terminal/truncated `Timestep` is returned as-is instead of silently resetting. Useful for eval loops, manual episode stepping, and debugging — a common ask for gym-style envs.
- 6 new `FourRooms` size registrations: `Navix-FourRooms-7x7-v0` through `-17x17-v0` (odd sizes, matching the room-splitting geometry), following the existing registration pattern exactly.

## Commits (cherry-picked, authorship preserved)
- `feat: disable autoreset on flag` (72e1eb5)
- `feat: four rooms env of different sizes` (c3bb465)
- `fix: odd numbers for four rooms` (6bc8170)
- `fix: jnp-native and/not` (cf75086)
- `fix: remove duplicate kwargs passing` (e990f74)

## Test plan
- [ ] CI passes (Test matrix, Compliance)
- [ ] Confirm `disable_autoreset=True` correctly keeps returning a terminal `Timestep` past `max_steps` instead of auto-resetting
- [ ] Confirm all 6 new `Navix-FourRooms-*-v0` IDs resolve via `nx.make(...)`

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01Be5vnY6tpvoTAwWAFN5mtH

PR #121, opened 2026-08-14
fix: FourRooms partition wall hardcoded to position 9

epignatelli:
## Summary
Addresses review findings from Claude on two open/merged PRs:

### 1. FourRooms partition wall hardcoded to position 9 ([review on #117](https://github.com/epignatelli/navix/pull/117#issuecomment-5299110931))
`FourRooms._reset()` hardcodes the partition wall position as `9` - correct only for the original `19x19` default. For the 6 smaller sizes #117 registered (`7x7` through `17x17`), position `9` is out of bounds or badly off-centre, so those rooms likely weren't actually split into four quadrants.

Traced the exact bug: `vertical_wall(grid, X, openings)`'s `X` is used internally as the fixed **column** for a wall spanning the full height, and `horizontal_wall(grid, Y, openings)`'s `Y` is the fixed **row** for a wall spanning the full width - the opposite of what the parameter names (`row_idx` / `col_idx`) suggest. Since the `19x19` default is square, `9 == 19 // 2` regardless of which dimension it's meant to track, which is exactly why this was never caught. Fixed as `vertical_wall(grid, self.width // 2, ...)` and `horizontal_wall(grid, self.height // 2, ...)`.

### 2. Two regressions in `open()` from #118 ([review on #118](https://github.com/epignatelli/navix/pull/118), already merged)

1. **Dtype regression**: writing back `is_open` (always bool) instead of `doors.open` (whatever dtype the constructing environment used - int for KeyCorridor/GoToDoor, bool for DoorKey) meant every door's `open` field silently became bool after toggling. `toggle()` is dispatched via `jax.lax.switch` alongside sibling actions that don't touch `doors.open` at all, and `lax.switch` requires identical output dtypes across every branch - so this raised a trace-time `TypeError` for any env with an int-typed door. DoorKey's tests didn't catch it because DoorKey already uses bool.
2. **Key-destruction regression**: `can_unlock` didn't check whether the door was already open, only that it was found/locked/key-matching. KeyCorridor's target door is constructed already open (`open=jnp.asarray(2)`) while still locked (`requires=key_id`) - a pre-existing quirk of its design. Standing in front of it with the matching key destroyed the key for nothing, since `do_open` was `False` (already open) but the pocket got cleared anyway. Before #118 this couldn't happen (the old code was a provable no-op), so this was a new regression, not a pre-existing bug.

Fix: `is_open` is now used only for the boolean predicate (`do_open`), never written back - the `replace()` call uses `doors.open` directly, so dtype is preserved via `jnp.where`'s weak-type promotion, matching pre-#118 behaviour. Key consumption is now gated on `do_open & locked & key_match` (an actual closed -> open transition on a locked, matching door), not the weaker `can_unlock`.

## Test plan
- [ ] CI passes
- `tests/test_environments.py::test_fourrooms_sizes_partition_walls_are_centred` - new; asserts every partition-wall cell sits on the `row=height//2` / `col=width//2` cross and stays in-bounds, for all 6 new sizes. Fails against the old hardcoded `9`, passes with this fix.
- `tests/test_environments.py::test_disable_autoreset` - new; closes another review-flagged coverage gap. Confirms `step_type` stays terminal past `max_steps` when enabled, and autoresets by default when not.
- `tests/test_actions.py::test_open_preserves_door_dtype` - new; an int-dtype door (KeyCorridor/GoToDoor-style) stays int-dtype after `open()`.
- `tests/test_actions.py::test_open_does_not_consume_key_for_already_open_door` - new; a door that's already open-but-locked (KeyCorridor's actual construction) leaves the key untouched.

## Not addressed
The review on #117 also suggested adding the 6 new size IDs to `docs/home/environments.md`. Skipped: that table is a strict MiniGrid-ID -> NAVIX-ID mapping, and these sizes have no MiniGrid equivalent, so adding them would misleadingly imply MiniGrid has these sizes too.

### DIFF
--- a/navix/actions.py
+++ b/navix/actions.py
@@ -269,21 +269,30 @@ def open(state: State) -> State:
     is_open = jnp.asarray(doors.open, dtype=jnp.bool_)
     locked = doors.requires != -1
     key_match = player.pocket == doors.requires
-    can_unlock = door_found & locked & key_match
-    can_open = door_found & (can_unlock | ~locked)
+    can_open = door_found & (key_match | ~locked)
 
-    # update doors if closed and can_open
+    # update doors if closed and can_open. Note: `is_open` is only used for
+    # the boolean predicate above - the write-back uses `doors.open` itself,
+    # not `is_open`, so `doors.open`'s original dtype (int or bool,
+    # depending on which environment constructed the door) is preserved via
+    # jnp.where's weak-type promotion. Writing back `is_open` instead would
+    # silently force every door to bool, which breaks jax.lax.switch's
+    # requirement that every action branch produce identical output dtypes.
     do_open = ~is_open & can_open
-    open = jnp.where(do_open, True, is_open)
+    open = jnp.where(do_open, True, doors.open)
     requires = jnp.where(do_open, -1, doors.requires)
     doors = doors.replace(open=open, requires=requires)
 
-    # remove key from player's pocket, but only when it was actually used to
-    # unlock a locked door - not merely because the door in front happened to
-    # be open or unlocked already
-    pocket = jnp.where(jnp.any(can_unlock), EMPTY_POCKET_ID, player.pocket)
+    # remove key from player's pocket, but only when this action actually
+    # unlocked a previously-closed, locked door with a matching key - not
+    # merely because the door in front happened to already be open (some
+    # environments, e.g. KeyCorridor, construct a door that is already open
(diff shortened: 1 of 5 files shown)


epignatelli:
@claude can you have a look and review please?

epignatelli:
@claude can you check if comments have been resolved? We keep all these fixes in one PR.

epignatelli:
@claude approve and merge

PR #123, opened 2026-08-15
fix: categorical_first_person no longer inverts entity transparency

### LINKED ISSUE #112: categorical_first_person and rgb_first_person disagree on entity occlusion (`~transparent`)
### Summary

`categorical_first_person` and `rgb_first_person` build the same view-cone
transparency map with opposite conventions, so they disagree about whether a
placed entity occludes vision.

In `navix/observations.py`:

```python
# categorical_first_person
transparency_map = transparency_map.at[tuple(positions.T)].set(~transparent)

# rgb_first_person
transparency_map = transparency_map.at[tuple(positions.T)].set(transparent)
```

One of these is wrong, and the map's own construction says which:

```python
transparency_map = jnp.where(state.grid == 0, 1, 0)
```

Floor (`grid == 0`) is written as `1` and walls (`-1`) as `0`, so the
convention is **1 = light passes, 0 = blocks** — which is also what `view_cone`
assumes, since its diffusion multiplies by the map and a `0` stops propagation.

`get_transparency()` means "light passes through this entity". Writing it
un-negated (`rgb_first_person`) matches the convention. Writing `~transparent`
(`categorical_first_person`) inverts it: an entity that is transparent is
recorded as opaque, and vice versa.

### Consequence

Under `categorical_first_person`, every placed entity, i.e. balls, keys, boxes,
blocks line of sight, and a closed door does not. Under `rgb_first_person` on
the same state, the opposite occurs. Agents trained on the two observation
types therefore see structurally different worlds.

### Comparison with MiniGrid

MiniGrid, which this repo already depends on in `requirements_test.txt`, treats
only walls as occluders: `Ball.see_behind()`, `Box.see_behind()` and
`Key.see_behind()` all return `True`; only `Wall` returns `False` (verified
against `minigrid` 3.1.0). That matches `rgb_first_person`, not
`categorical_first_person`.

### Why it has gone unnoticed

The observation tests do not assert on this. `test_rgb_first_person` builds a
navix observation and a MiniGrid one and then returns without comparing them,
and `test_categorical_first_person` ends in `print(obs)`. So nothing currently
fails if the two observation functions disagree, or if either drifts from
MiniGrid.

### Questions

1. Which convention is intended? The code-level evidence points at the
   un-negated form, but the negation is old enough that it may be deliberate.
2. If the un-negated form is correct, the fix is one character, but it changes
   every `categorical_first_person` observation in the library, so anyone with
   trained agents or cached datasets would be affected. That probably deserves
   a release note rather than a quiet patch.

Happy to open a PR for whichever direction you prefer, and to make
`test_rgb_first_person` actually assert parity with MiniGrid so this class of
drift is caught in future.

Related: [PR111](https://github.com/epignatelli/navix/pull/111) fixes a
separate, purely-type-level problem in the same area (`Door.walkable`/
`Door.transparent` returning `open`'s dtype rather than bool). This issue
is independent of that one though, as the inversion is present regardless of
dtype.

N.B.: written with AI assistance for the investigation and drafting, but all
verifications are my own.

epignatelli:
## Summary
Fixes #112.

\`categorical_first_person\` and \`rgb_first_person\` built the same view-cone transparency map with opposite conventions:

\`\`\`python
# categorical_first_person
transparency_map = transparency_map.at[tuple(positions.T)].set(~transparent)

# rgb_first_person
transparency_map = transparency_map.at[tuple(positions.T)].set(transparent)
\`\`\`

The map's own construction (\`jnp.where(state.grid == 0, 1, 0)\`) writes floor as \`1\` and walls as \`0\` — i.e. **1 = light passes, 0 = blocks** — matching \`view_cone\`'s diffusion, which multiplies by the map and stops propagation at \`0\`. \`Entity.get_transparency()\` means "light passes through this entity", so writing it un-negated (\`rgb_first_person\`'s way) matches the convention; negating it (\`categorical_first_person\`'s way) inverted every entity's occlusion — a closed door blocked vision under \`rgb_first_person\` but *let it through* under \`categorical_first_person\`, and vice versa for open doors/other entities.

Confirmed against MiniGrid 3.1.0 per the issue: only \`Wall.see_behind()\` returns \`False\`; \`Ball\`/\`Box\`/\`Key\` all return \`True\` — matching the un-negated convention, not the negated one.

## Test plan
- [x] CI passes
- \`tests/test_observations.py::test_categorical_first_person_transparency_matches_rgb\` — new; places a transparent \`Key\` directly in front of the player with a \`Goal\` one cell further, asserts the \`Goal\` is visible through the \`Key\`. Fails against the old negated code, passes with the fix.

## Note on impact
Per the issue: this changes every \`categorical_first_person\` observation in the library (entities now correctly let vision through instead of blocking it, and vice versa) — anyone with agents trained on this observation type will see a behavior change. Worth calling out in release notes.

### DIFF
--- a/navix/observations.py
+++ b/navix/observations.py
@@ -94,7 +94,7 @@ def categorical_first_person(state: State) -> Array:
     transparency_map = jnp.where(state.grid == 0, 1, 0)
     positions = state.get_positions()
     transparent = state.get_transparency()
-    transparency_map = transparency_map.at[tuple(positions.T)].set(~transparent)
+    transparency_map = transparency_map.at[tuple(positions.T)].set(transparent)
 
     # apply view mask
     player = state.get_player()
--- a/tests/test_observations.py
+++ b/tests/test_observations.py
@@ -5,7 +5,7 @@ import jax.numpy as jnp
 
 import navix as nx
 from navix.states import State
-from navix.entities import Entities, Player, Goal, Key, Door
+from navix.entities import Entities, EntityIds, Player, Goal, Key, Door
 from navix.components import EMPTY_POCKET_ID
 from navix.rendering.cache import RenderingCache, TILE_SIZE
 from navix.rendering.registry import SPRITES_REGISTRY, PALETTE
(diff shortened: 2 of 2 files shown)


epignatelli:
@claude review

PR #129, opened 2026-08-15
perf: skip per-step log indexing for steps log_frequency discards

### LINKED ISSUE #60: Speed up logging

epignatelli:
## Summary
Fixes #60 ("Speed up logging").

Profiled a real PPO run (`Navix-Empty-8x8-v0`, 4 seeds, A100 80GB) rather than guessing - full numbers below.

### Before
- `do_log=False` (4 seeds): 0.04s
- 1 seed, `do_log=True`, under cProfile: **28.4s**, of which **17.4s (61%)** was `jax._src.compiler...backend_compile_and_load` - XLA recompiling from scratch, not wandb network I/O.
- 4 seeds, `do_log=True`, sequential (current code): **28.5s total** - essentially the *same* wall-clock time as a single seed, directly contradicting `Experiment.run`'s own "linear in the number of seeds" docstring warning.

### Root cause
`lengths[mask]` / `returns[mask]` in `Agent.log()` boolean-index a variable number of completed episodes per training update. That output shape depends on the mask's population count, which differs almost every step, so JAX must compile a fresh XLA program for every distinct count it hasn't seen before.

### Fix (commit 1-2)
Compute the same means via masked sum/count (`jnp.where` + `jnp.sum`, factored out as `masked_mean()`) instead of boolean-indexing a variable-length subset. Output shape is always fixed regardless of how many entries are masked, so XLA compiles it once, ever, and reuses it.

Verified directly (not just inferred from cProfile's top-N): with `JAX_LOG_COMPILES=1`, every distinct compile signature in the fixed code appears **exactly once** across a 97-step run. The old code recompiled `gather`/`equal` separately for *every* distinct episode-count shape (confirmed: 10+ separate `gather` compiles, one per shape, in the same run).

### After
- 1 seed, `do_log=True`: **5.99s (4.7x faster)**
- 4 seeds, `do_log=True`, sequential: **24.7s total (~6.2s/seed)** - now genuinely linear, matching the docstring, since compilation is no longer a confound.

### Concurrency, tried and reverted (commits 3, 6, 8 - see below)
An earlier version of this PR made `Experiment.run()`'s (and `run_hparam_search()`'s) per-seed logging loop concurrent, first via `ThreadPoolExecutor`, then via `ProcessPoolExecutor(spawn)` after the threaded version turned out to corrupt runs. Both were reverted - see "Revert: back to sequential logging" below for the full story and real measurements. **Net effect: logging is sequential, same as before this PR, just using the `log_to_wandb` rename.**

### Plotting (commit 4-5)
Adds `navix.plotting`, complementing the `do_log=False` + read `logs` directly workflow the maintainer's own comment on #60 already recommends, with a way to actually look at the results without wandb:

- `MANDATORY_METRICS`: a fixed, reasoned-about set (`perf/returns`, `perf/success_rate`, `perf/episode_length`, `iter/fps`, `iter/wall_time`) identical across every agent, so plots stay comparable across algorithms - laying groundwork for the leaderboard proposal (#130). Each metric is derivable purely from the (state, action, reward, done) interaction stream plus wall-clock time, so the set stays meaningful for algorithms navix doesn't ship itself.
- Diagnostic (algorithm-specific) metrics are intentionally *not* defined anywhere in navix core - a submitted algorithm won't necessarily have navix-specific code to declare them. `plot_metrics`/`plot_dashboard` both just accept an arbitrary `metrics` dict, so a caller (e.g. a future leaderboard's own algorithm -> diagnostic-keys mapping) can merge its own metrics with `MANDATORY_METRICS` and pass the result in.
- `plot_metric`/`plot_metrics` produce individual standalone figures; `plot_dashboard` produces one combined figure, one panel per requested metric.
- `derive_scalar_metrics()` reuses the exact `masked_mean()` helper from the fix above to compute `perf/*` for an entire logs history at once.
- Adds `matplotlib` as an official dependency, lazy-imported so `import navix` itself doesn't require it.

### Review fixes (commit 6)
- Fixed a `ThreadPoolExecutor(max_workers=...)` crash on empty `seeds`/`pop_size=0` (moot after the revert below, since there's no executor anymore, but real at the time).
- Removed a stale `DIAGNOSTIC_METRICS` docstring reference in `plot_metrics` (that attribute was removed from `PPO` in favor of a plain `metrics` param, see commit 5).
- Added `tests/test_experiment.py` (rewritten twice more since, see the revert below).

### Rename: log/log_on_train_end/do_log -> log_to_wandb (commit 7)
`Agent.log`, `Agent.log_on_train_end`, and `Experiment.run(do_log=)` only ever sent data to wandb, but nothing in the names said so - misleading now that `navix.plotting` (commit 4-5) exists as a second, local-only way to look at a run. Renamed to `Agent.log_to_wandb`, `Agent.log_to_wandb_on_train_end`, `Experiment.run(log_to_wandb=)`.

Old names/kwarg kept as deprecated shims (`DeprecationWarning`, delegate to the new ones), not removed - existing callers keep working. All internal navix call sites (`ppo.py`'s debug callback, `experiment.py`, `benchmarks/navix_.py`) use the new names directly, so normal usage never triggers the warning. The shims are plain Python (`warnings.warn` + delegate) reached outside JAX tracing - `ppo.py`'s `jax.debug.callback` points at `log_to_wandb` directly rather than through the deprecated alias - so this doesn't affect compilation.

Docstrings (`Agent`, `Experiment.run`, `navix.plotting`'s module docstring) now spell out the two strategies explicitly: `log_to_wandb=True` (wandb, slower, real network I/O) vs. `log_to_wandb=False` + `navix.plotting` (local matplotlib, no network calls).

### Revert: back to sequential logging (commit 8)
CI broke on this branch after the rename: `examples/hparam_search.py` (10-way hparam search) hit `wandb.errors.errors.UsageError: No API key configured` under `wandb offline` - concurrent `wandb.init()` calls from multiple threads raced into a real login attempt. Serializing just `wandb.init()` with a lock fixed that race, but re-running the real reproduction on GPU then hit a second, worse bug: `Run (...) is finished. The call to \`log\` will be ignored.` - a run got corrupted by another thread's concurrent `.log()`/`.finish()`, even with each thread using its own explicit `Run` object.

Tried replacing threads with `ProcessPoolExecutor(mp_context=multiprocessing.get_context("spawn"))` instead (`spawn`, not the Linux-default `fork`, to stay safe w.r.t. JAX/CUDA - forking a process after CUDA is initialized in the parent is undefined behavior regardless of timing, since every thread but the calling one just vanishes in the child, taking whatever locks it held with it; also hid the GPU from workers entirely via `CUDA_VISIBLE_DEVICES=""` as a belt-and-braces guard). This fixed correctness, verified against the real `examples/hparam_search.py`/`examples/ppo.py` reproductions under `wandb offline`. But then measured whether it was actually still a speed win, with a controlled sweep (same trained `logs`, sequential vs. multiprocess, 1/2/4/8/16 seeds) on a real GPU:

| seeds | sequential | multiprocess | verdict |
|---|---|---|---|
| 1 | 4.25s | 11.71s | multiprocess **2.8x slower** |
| 2 | 6.75s | 12.13s | multiprocess **1.8x slower** |
| 4 | 10.56s | 14.86s | multiprocess **1.4x slower** |
| 8 | 24.70s | 20.42s | multiprocess **1.2x faster** |
| 16 | 36.47s | **crashed** | `MemoryError: std::bad_alloc` + `Failed to launch ptxas` |

Multiprocessing only wins at 8+ concurrent seeds (each worker pays a real cost re-importing the entire jax/flax/distrax/tensorflow_probability/wandb stack from a cold interpreter), is worse than sequential below that, and at 16 workers isn't even reliable - it exhausted host memory. Since `seeds=(0,)` is the default and most usage is a handful of seeds, "always multiprocess" made the common case worse in exchange for a win nobody's default configuration would ever see, plus a real crash risk at higher counts nobody had capped for.

Reverted to a plain sequential `for` loop - the same shape `Experiment.run`/`run_hparam_search` had before this PR touched logging at all, now just using the `log_to_wandb` rename. `run()`'s docstring still calls out that this is real, roughly-linear-in-seeds cost, since that's genuinely true and worth knowing before choosing `log_to_wandb=True`.

(Also checked [`mila-iqia/parallel_wandb`](https://github.com/mila-iqia/parallel_wandb), a reference implementation built specifically for many-wandb-runs-from-one-JAX-process: it also just uses a sequential `for` loop for `wandb.init()`/`wandb.log()`. Its "parallel" refers to `jax.vmap` training multiple runs together, with per-step logging via `io_callback` from inside the training loop, not to thread/process concurrency of the logging calls themselves - a materially different architecture from what this PR does (train fully, then log after). Its use of `wandb.init(reinit="create_new")` suggests the threaded corruption bug above may have been fixable with that flag rather than a fundamental thread-safety wall, but since even that reference implementation doesn't attempt cross-thread/process concurrency for init/log/finish, there's no evidence a threaded approach would have been worth chasing further here.)

## Test plan
- [x] CI passes
- `tests/test_agents.py` - log_frequency behavior + masked-mean correctness vs plain numpy boolean-indexing
- `tests/test_plotting.py` - `derive_scalar_metrics` vs plain numpy reference, dashboard layout has exactly one axes per requested metric
- `tests/test_experiment.py` - every seed gets its own `Run`, logged and finished exactly once; `do_log` deprecation; empty-seeds edge case doesn't crash
- Ran the full test suite (45 passed) plus `examples/hparam_search.py` and `examples/ppo.py` directly under `wandb offline` (the exact scenario that broke CI) on a real GPU (A100) at each step of this PR, since CI here only validates on the hardware this repo's runners have

### DIFF
--- a/benchmarks/navix_.py
+++ b/benchmarks/navix_.py
@@ -61,7 +61,7 @@ if __name__ == "__main__":
         seeds=tuple(range(args.seeds_offset, args.seeds_offset + args.n_seeds)),
         group="navix",
     )
-    train_state, logs = experiment.run(do_log=False)
+    train_state, logs = experiment.run(log_to_wandb=False)
 
     print("Logging final results to wandb...")
     start_time = time.time()
@@ -69,7 +69,7 @@ if __name__ == "__main__":
     logs_avg = jax.tree.map(lambda x: x.mean(axis=0), logs)
     config = {**vars(experiment), **asdict(agent.hparams)}
     wandb.init(project=experiment.name, config=config, group=experiment.group)
-    agent.log_on_train_end(logs_avg)
+    agent.log_to_wandb_on_train_end(logs_avg)
     wandb.finish()
     logging_time = time.time() - start_time
     print(f"Logging time cost: {logging_time}")
--- a/navix/__init__.py
+++ b/navix/__init__.py
@@ -32,6 +32,7 @@ from . import (
     transitions,
     events,
     agents,
+    plotting,
 )
 
 from .environments.registry import make, register_env, registry
--- a/navix/agents/agent.py
+++ b/navix/agents/agent.py
@@ -39,19 +39,30 @@ class Agent(struct.PyTreeNode):
 
         if "done_mask" in logs:
             mask = jnp.asarray(logs.pop("done_mask"), dtype=jnp.bool)  # (T, N)
+            # Masked mean via sum/count instead of boolean-indexing
+            # (`lengths[mask]`) - mask[mask] has a *dynamic* shape (however
+            # many episodes finished this update, which varies every call),
+            # so JAX must recompile a fresh XLA program for every distinct
+            # count it hasn't seen before. Profiling showed this was ~60%
+            # of total per-call logging time. jnp.where + jnp.sum keeps the
+            # shape fixed at (T, N) -> scalar regardless of how many
+            # entries are masked, so XLA compiles it once and reuses it.
+            mask_count = jnp.sum(mask)
             # log episode length
             if "lengths" in logs:
                 lengths: jax.Array = logs.pop("lengths")  # (T, N)
-                episode_lengths = lengths[mask]  # (K,)
-                logs["perf/episode_length"] = jnp.mean(episode_lengths)
+                logs["perf/episode_length"] = (
+                    jnp.sum(jnp.where(mask, lengths, 0)) / mask_count
+                )
                 msg += f", Length: {logs['perf/episode_length']}"
 
             # log returns
             if "returns" in logs:
                 returns = logs.pop("returns")  # (T, N)
-                final_returns = returns[mask]  # (K,)
-                logs["perf/returns"] = jnp.mean(final_returns)
-                logs["perf/success_rate"] = jnp.mean(final_returns == 1.0)
+                logs["perf/returns"] = jnp.sum(jnp.where(mask, returns, 0)) / mask_count
+                logs["perf/success_rate"] = (
+                    jnp.sum(jnp.where(mask, returns == 1.0, False)) / mask_count
+                )
                 msg += f", Returns: {logs['perf/returns']}, Success Rate: {logs['perf/success_rate']}"
 
         msg += f", Logging time cost: {time.time() - start_time}"
--- a/navix/agents/agent.py
+++ b/navix/agents/agent.py
@@ -60,6 +60,16 @@ class Agent(struct.PyTreeNode):
     def log_on_train_end(self, logs):
         print(jax.tree.map(lambda x: x.shape, logs))
         len_logs = len(logs["iter/updates"])
+        updates = logs["iter/updates"]
         for step in range(len_logs):
+            # skip steps self.log() would discard anyway (see the
+            # log_frequency check below), before paying for the
+            # device-to-host transfer of indexing into every array in
+            # `logs` - wandb.log() and this per-step tree indexing were
+            # previously done unconditionally for every recorded step,
+            # which is why disabling wandb logging entirely (do_log=False)
+            # was so much faster than leaving it on
+            if updates[step] % self.hparams.log_frequency != 0:
+                continue
             step_logs = {k: jax.tree.map(lambda x: x[step], v) for k, v in logs.items()}
             self.log(step_logs)
--- a/navix/agents/agent.py
+++ b/navix/agents/agent.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
 import time
+import warnings
 from typing import Dict, Tuple
 
 import numpy as np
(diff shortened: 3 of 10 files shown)


epignatelli:
@claude please review

epignatelli:
Thanks for the review - responding point by point (fixed in `56f8a47`):

1. **`ThreadPoolExecutor` crashes on empty input** - confirmed real, fixed: `max_workers` now clamps to `max(len(self.seeds), 1)` / `max(len_search_set, 1)` in both `run()` and `run_hparam_search()`, restoring the prior silent-no-op behavior for empty `seeds`/`pop_size=0`. Added `tests/test_experiment.py::test_experiment_run_with_no_seeds_does_not_crash` to cover it.

2. **Stale PR description / stale `DIAGNOSTIC_METRICS` docstring** - both fixed. Updated the PR body to reflect the actual final state (generic `metrics` dict, not `PPO.DIAGNOSTIC_METRICS`, not a 2-row mandatory+diagnostic layout), and fixed the stale docstring line in `plot_metrics` (`navix/plotting.py`).

3. **Concurrent `wandb.init()` thread-safety is unverified** - agreed, this is a real open question rather than something I can rule out from here. Added a code comment at the `ThreadPoolExecutor` call site in `navix/experiment.py` flagging it explicitly as the first place to look if runs are ever misattributed/corrupted rather than crashing outright. Not fixing further in this PR (would mean multiprocessing, which has its own real JAX/CUDA fork-safety cost, discussed in the PR description) - happy to open a follow-up issue if you'd rather track it separately.

4. **No test coverage for the concurrency path** - added `tests/test_experiment.py`: a smoke test with a fake `Agent`/`Run` (no real training, no real wandb backend) asserting every seed gets its own `Run`, logged and finished exactly once, plus the zero-seeds case from (1).

5. **`iter/wall_time` renders as a flat line** - correct, and intentional: it's the total elapsed time for the whole run, broadcast across updates the same way the pre-existing `iter/fps` already is (there's no way to get true per-step wall-clock without host-callback timing inside the `jax.lax.scan`, which would reintroduce the exact per-step overhead #60 fixed). As a flat line it's still directly useful for cross-algorithm "cost to train" comparison on the leaderboard (#130), which is what it's for - not meant to be a training-progress curve like the other panels. Not changing this, but appreciate the sanity check.

All of `tests/test_agents.py`, `tests/test_plotting.py`, `tests/test_experiment.py` (10 tests) plus the full suite (42 passed) verified on a real GPU (A100) after these fixes.

epignatelli:
@claude can you please review again after the changes?

epignatelli:
@claude last review before merge please

epignatelli:
@claude please review again

Issue #134, opened 2026-08-15
Master: port useful work from other branches and forks


epignatelli:
## Purpose

Tracking issue for reviewing unmerged work sitting on other branches/forks and deciding what's worth porting into `main`. This issue is the index - each row in the table below gets its own issue/PR when actually worked on, linked back here.

## 1. Sources with interesting code to port

- **[`arahosu/navix_minigrid`](https://github.com/arahosu/navix_minigrid)** (external fork) - branches `fix/crossings-monotone-path`, `fix/ghost-key-observations`, `notebooks`. A systematic, reference-validated bug hunt (diffed against Farama MiniGrid 3.0.0 as ground truth), documented in `notebooks/VALIDATION_REPORT.md` on the fork's `notebooks` branch. Two bugs already have fix branches + regression tests ready to port; one needs a design decision before any fix is written; one is likely not worth pursuing (the report also caught a *previously-documented* "fix" that turned out to be wrong - worth reading in full before acting).
- **`refactor-events`** (our own branch, unmerged, June 2024, unfinished/WIP). Redesigns the events system to support multiple simultaneous events per step instead of one scalar `Event` field per type - structurally relevant to bugs like #91/PR #125.
- **`speedup` / `neurips`** (your active research branches - `neurips`'s tip is `speedup`'s tip plus one more commit, so there's nothing to reconcile *between* them; review only, don't touch either directly). Paper/rebuttal code: throughput benchmarks, the `baselines/rejax` training script, a full DreamerV3-style agent, IQN/PQN baseline results, and a few source-code changes worth reconciling against what's already in `main`.

### Ruled out (investigated, not worth porting)

- `ppo-rnn` (recurrent PPO agent, June 2024) - explicitly out of scope per maintainer.
- `codex/investigate-github-issue-#100` (PR #93) - deliberately closed; a different approach (`default_factory`) was chosen instead and already shipped.
- `rejax` branch's `examples/rejax_.py` - superseded by the much more complete `baselines/rejax/` already in `main`.
- `website` branch - trivial pre-mkdocs tweaks, docs infrastructure has since moved to mkdocs + gh-pages.
- `benchmarks` branch's throughput scripts - not redundant with anything in `main`, but 2+ years stale against the current API; superseded in spirit by what's tracked under speedup/neurips above.
[...]

PR #145, opened 2026-08-15
fix: picked-up entities no longer ghost into categorical/symbolic obs

### LINKED ISSUE #135: fix: ghost key wrap-around in categorical/symbolic observations
Part of #134.

## Bug

When an entity is picked up, its position moves to `DISCARD_PILE_COORDS = (0, -1)`. `observations.py`'s `categorical()`/`symbolic()` don't validate that position before writing the entity tag/symbol into the observation array, so Python's negative-indexing semantics wrap it into a *valid* cell instead of being a no-op:

- `categorical()`: `flat_grid.at[-1].set(KEY_TAG)` writes to flat index `H*W-1` → bottom-right `(H-1, W-1)`.
- `symbolic()`: `obs.at[(0, -1)].set(KEY_SYMBOL)` writes to `(0, W-1)` → top-right corner.

Result: a phantom duplicate of the picked-up item appears elsewhere in the observation. In MiniGrid's reference implementation, picked-up objects are removed from the grid entirely (held by `agent.carrying`) - the encoded observation has no trace of them.

## Verification

Reference-validated (DoorKey-8x8, seed 0, after `[rotate_cw, forward, pickup]`):
- MiniGrid full-grid encode: no key in grid. ✅
- Navix `main` `categorical()[7, 7] = 5` (KEY tag). ❌ ghost key
- Navix `main` `symbolic()[0, 7] = [5, 4, 0]` (KEY symbol). ❌

## Fix (already written, ready to port)

Source: [`arahosu/navix_minigrid:fix/ghost-key-observations`](https://github.com/arahosu/navix_minigrid/tree/fix/ghost-key-observations), by Joonsu Gha ([@arahosu](https://github.com/arahosu)/[joogha@ethz.ch]).

Adds a validity mask to both functions - writes whose flat index (categorical) or 2-D position (symbolic) fall outside the grid become no-ops. Restores exactly 1 cell per observation to its pre-pickup value; verified bit-identical to the unmodified obs everywhere else. ~10 lines, `navix/observations.py` + a regression test in `tests/test_issues.py`.

Full validation writeup: `notebooks/bug_c_ghost_key.ipynb` and `notebooks/VALIDATION_REPORT.md` on the fork's `notebooks` branch.

## Next steps

Port the fix branch as a PR, crediting the original author. Smallest, most self-contained item in #134 - good first port.

epignatelli:
## Summary
Fixes #135 and #146 (item 1 of the port-tracking master issue #134).

When an entity is picked up, its position moves to `DISCARD_PILE_COORDS = (0, -1)`. `categorical()`/`symbolic()` scatter-write the entity's tag/symbol at its position without checking whether that position is actually on the grid - JAX's negative-index wraparound then writes it into a real cell instead of dropping it, producing a "ghost" duplicate of the picked-up item: `categorical()`'s flat index `-1` wraps to the bottom-right corner, `symbolic()`'s `(0, -1)` wraps to the top-right corner.

`rgb()`/`rgb_first_person()` already solve a version of this problem: `state.cache.patches` has one extra "sink" cell that a discarded entity's wraparound lands in, then `patches[:DISCARD_PILE_IDX]` drops it. An initial version of this PR mirrored that pattern for `categorical()`/`symbolic()`, but there's a simpler fix that doesn't need an extra allocated cell at all: `.at[].set(..., mode="drop")` already exists specifically to discard out-of-bounds writes - it just doesn't trigger on negative indices by default, since JAX normalizes those via numpy wraparound semantics *before* the bounds check (verified empirically: `.at[-1].set(x, mode="drop")` on a length-N array still wraps to index N-1, it does not drop). Pushing invalid indices/positions to be explicitly `>= size` first, then using `mode="drop"`, avoids needing to resize/pad the array and slice a sink cell back off afterward:
- `categorical()`: remaps any negative flat index to `num_cells` (one past the end) before the scatter.
- `symbolic()`: remaps any off-grid `(row, col)` to `(H, W)` before the scatter (checked via a proper `0 <= row < H and 0 <= col < W` validity check, not just the column, so it's not narrowly tied to the exact `(0, -1)` sentinel).

## Verification

- `tests/test_issues.py::test_135` - a direct `State` construction with a picked-up key at `DISCARD_PILE_COORDS`. Fails on pre-fix code (`categorical()`'s bottom-right corner reads the KEY tag instead of the wall); passes after the fix.
- A real `Navix-DoorKey-8x8-v0` rollout: picked up the key (`[rotate_cw, forward, pickup]`), checked both observation functions directly. Pre-fix, both ghost the key into their respective corners; post-fix both correctly show the wall.
- Explicitly re-checked under both `jax.jit` and `jax.vmap` (observations get called from inside jitted/vmapped training, not just eagerly) - both clean.
- Full test suite (46 passed) and `examples/*.py` (both exit cleanly) verified on a real GPU.

## categorical_first_person (added per review, #146)

Review feedback caught that `categorical_first_person` has the identical bug: `transparency_map`/`state.grid` are scattered via raw entity positions without checking they're on-grid, same failure mode as `categorical()`/`symbolic()`. Originally filed as a separate follow-up issue (#146) since it was out of #135's original scope, but folded into this PR instead since it's the same fix pattern and touches the same file.

Fixed with the same `mode="drop"` + explicit out-of-bounds remap. `tests/test_issues.py::test_146` uses a differential check (compare against an otherwise-identical state with no key entity at all) rather than hand-computing which cell of `crop()`'s pad/roll/rotate/slice output the wraparound lands in - empirically verified this actually catches the bug (at every radius from 3 to 20 tried) before finalizing the test, so it's not accidentally vacuous.

## Attribution

`categorical()`/`symbolic()` ported from [`arahosu/navix_minigrid:fix/ghost-key-observations`](https://github.com/arahosu/navix_minigrid/tree/fix/ghost-key-observations), found and reference-validated against Farama MiniGrid by Joonsu Gha ([@arahosu](https://github.com/arahosu)) - see the fork's `notebooks/VALIDATION_REPORT.md` (linked from #135) for the full investigation, including a separate case where a *different*, previously-documented "fix" for a different bug turned out to be wrong on closer inspection (independently verified by viewing the notebook's comparison images directly, not just the text summary). `categorical_first_person`'s fix (#146) is this repo's own follow-up, same bug class, not part of the fork's original report.

## Test plan
- [ ] CI passes
- `tests/test_issues.py::test_135`, `test_146` - direct repro + regression tests
- Full suite (47 passed) + examples verified on real GPU (see above)

### DIFF
--- a/navix/observations.py
+++ b/navix/observations.py
@@ -101,7 +101,19 @@ def categorical_first_person(state: State) -> Array:
     transparency_map = jnp.where(state.grid == 0, 1, 0)
     positions = state.get_positions()
     transparent = state.get_transparency()
-    transparency_map = transparency_map.at[tuple(positions.T)].set(transparent)
+    # a picked-up entity's position (DISCARD_PILE_COORDS = (0, -1)) is
+    # off-grid - .at[].set()'s default mode wraps negative components
+    # around (numpy semantics) rather than dropping them, silently
+    # overwriting a real cell (see categorical()/symbolic() for the
+    # same bug, fixed the same way). Push off-grid positions to be
+    # explicitly out of bounds first, so mode="drop" discards those
+    # writes instead.
+    H, W = state.grid.shape
+    row, col = positions[..., 0], positions[..., 1]
+    on_grid = (row >= 0) & (row < H) & (col >= 0) & (col < W)
+    row = jnp.where(on_grid, row, H)
+    col = jnp.where(on_grid, col, W)
+    transparency_map = transparency_map.at[row, col].set(transparent, mode="drop")
 
     # apply view mask
     player = state.get_player()
@@ -109,7 +121,7 @@ def categorical_first_person(state: State) -> Array:
 
     # get categorical representation
     tags = state.get_tags()
-    obs = state.grid.at[tuple(positions.T)].set(tags) * view
+    obs = state.grid.at[row, col].set(tags, mode="drop") * view
 
     # crop grid to agent's view
     obs = crop(obs, player.position, player.direction, RADIUS)
--- a/navix/observations.py
+++ b/navix/observations.py
@@ -73,15 +73,16 @@ def categorical(state: State) -> Array:
     indices = idx_from_coordinates(state.grid, state.get_positions())
     # get tags corresponding to the entities
     tags = state.get_tags()
-    # set tags on the flat set of patches. Append a discard-pile sink
-    # cell first: a picked-up entity's position (DISCARD_PILE_COORDS)
-    # maps to flat index -1, which .at[].set() wraps around to the
-    # *last* element rather than dropping - same pattern as rgb()'s
-    # cache.patches, so that wraparound lands on this dedicated extra
-    # cell instead of a real grid cell.
+    # set tags on the flat set of patches
     shape = state.grid.shape
-    grid = jnp.append(state.grid.reshape(-1), 0).at[indices].set(tags)
-    grid = grid[:DISCARD_PILE_IDX]  # drop the sink cell
+    num_cells = shape[0] * shape[1]
+    # a picked-up entity's position (DISCARD_PILE_COORDS) maps to flat
+    # index -1 - .at[].set()'s default mode wraps negative indices
+    # around (numpy semantics) rather than dropping them, silently
+    # overwriting a real cell. Push negative indices to be explicitly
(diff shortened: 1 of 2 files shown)


epignatelli:
@claude can you please add a review here. Make sure that the changes are not breaking the ful compilation of a training.

epignatelli:
@claude there's new commits

Issue #138, opened 2026-08-15
RGB first-person padding color diverges from MiniGrid (low priority)


epignatelli:
Part of #134.

## Background

An earlier, informally-documented "bug" claimed `grid.py::crop`'s slicing (`rotated.at[:diameter+1, ...]`) drops the top wall row of the egocentric view, and proposed changing it to `rotated.at[diameter-radius:diameter+radius+1, ...]`.

Joonsu Gha ([@arahosu](https://github.com/arahosu)) investigated this on the [`arahosu/navix_minigrid`](https://github.com/arahosu/navix_minigrid) fork (documented as "Bug A" in `notebooks/VALIDATION_REPORT.md`) and found **the proposed fix is wrong**: it moves the agent's position in the 7x7 egocentric view from bottom-center to center. MiniGrid actually puts the agent at bottom-center (row 6, col 3) - the *old*, unmodified Navix code already matches MiniGrid on agent placement. The proposed fix would have made this worse, not better.

## The actual (much smaller) issue

The visual symptom that prompted the original report ("missing top walls") is a different, real but minor bug: out-of-env/unseen cells in `rgb_first_person` render as `dark_cell_colour=0` (black), while MiniGrid renders the corresponding cells as wall-gray pixels.

## Recommendation: skip or defer

A one-line fix (`dark_cell_colour=0` → MiniGrid's wall RGB `(100, 100, 100)`) would bring the output visually closer, but pixel-perfect alignment is still blocked by `apply_minigrid_opacity` darkening visible walls relative to MiniGrid's brightness - so even the fix wouldn't fully close the gap. Marginal value relative to the discussion/review cost.

**Do not port the `crop()` slicing change** - it's actively wrong, not just unnecessary.

epignatelli:
Fixed in #147

PR #148, opened 2026-08-15
fix: first-person view only saw half its claimed radius, black padding

### LINKED ISSUE #147: First-person observations only see half their claimed view radius
## Bug

`categorical_first_person()` and `rgb_first_person()` both compute their visibility mask via:

```python
view = view_cone(transparency_map, player.position, RADIUS)
```

But `crop()` (used right after, to produce the actual `2*RADIUS+1` x `2*RADIUS+1` egocentric view) places the agent at the *bottom* row of the crop, meaning the top row of the view is `2*RADIUS` cells forward of the agent - not `RADIUS`. `view_cone`'s visibility diffusion, called with just `RADIUS`, only propagates half as far as the crop window actually extends. The result: the forward (far) half of every first-person view is permanently marked unseen/dark, regardless of whether real walls are there or not - not a fog-of-war effect, a hard cutoff at exactly the midpoint of the view.

## Verification

Confirmed directly, not just visually: in an open `Empty-16x16` room with the agent in the middle (no walls within `2*RADIUS` cells in any direction), `RADIUS=3`:

```
view_cone(transparency_map, origin, radius=3):  cropped mask ->
[[0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0]
 [1 1 1 1 1 1 1]
 [1 1 1 1 1 1 1]
 [1 1 1 1 1 1 1]
 [1 1 1 1 1 1 1]]

view_cone(transparency_map, origin, radius=6):  cropped mask ->
(entire 7x7 view visible - all 1s)
```

Doubling the radius passed to `view_cone` (i.e. using `2*RADIUS`, matching `crop()`'s actual forward reach) produces full visibility in the open room, matching a real MiniGrid render for the identical scenario (`MiniGrid-Empty-16x16-v0`, agent_view_size=7, same position/direction) side by side.

Rendered comparison (agent near a corner, so the effect is partly masked by legitimate off-map area, but still visible as an oversized black region) vs. the same scenario in the middle of an open room (where the top ~3 of 7 rows are black despite being genuinely in-bounds, open floor) both confirm this.

## Not the same as issue #138 (Bug A)

This was found while investigating #138 (RGB first-person padding color), but is a distinct, more significant bug: #138 is about the *color* used for legitimately out-of-map cells (cosmetic). This is about real, in-bounds, unobstructed cells being incorrectly marked invisible - a genuine visibility/field-of-view computation bug, not a rendering color choice. The fork's validation report that found #138 didn't catch this, since its test scenario (agent right next to a corner) confounds "off-map" with "unreached-by-diffusion," masking the effect.

## Scope

- `categorical_first_person()` (`navix/observations.py`) - affected, calls `view_cone(..., RADIUS)`.
- `rgb_first_person()` (`navix/observations.py`) - affected, same call pattern.
- `symbolic_first_person()` - **not** affected by this specific bug; it doesn't call `view_cone` at all, it crops the fully-observable `symbolic()` grid directly with no fog-of-war/visibility masking applied. That's a separate, likely pre-existing design characteristic (no partial-observability for the symbolic first-person view), not this bug - worth a separate look if partial observability is expected there too, but out of scope here.

## Next steps

Likely fix: pass `2 * RADIUS` (or equivalently `diameter`, already computed inside `crop()`) to `view_cone` in both affected functions, so the visibility diffusion reaches as far as the crop window actually shows. Needs verification that this doesn't meaningfully regress performance - `view_cone` switches between a fast direct method and a `lax.scan`-based method depending on radius relative to `MIN_SCAN_RADIUS = 10`; doubling the effective radius could cross that threshold for larger `RADIUS` settings and change which code path runs.

epignatelli:
## Summary
Fixes #147 and #138. Also fixes two more related issues found along the way (grid-line colour, and a int32 overflow in `view_cone` caught by @Near32's review) - see below.

### #147: view_cone radius mismatch (the real bug)
`categorical_first_person()`/`rgb_first_person()` computed their visibility mask via `view_cone(transparency_map, player.position, RADIUS)`. But `crop()` places the agent at the *bottom* row of the `2*RADIUS+1` view, so the far row is `2*RADIUS` cells forward of the agent, not `RADIUS`. `view_cone`'s diffusion only propagated half as far as the crop window actually extends, so the forward half of every first-person view was permanently marked unseen - a hard cutoff at exactly the midpoint, not fog-of-war, regardless of whether real walls were there.

Fixed by passing `RADIUS * 2` to `view_cone` in both functions, matching `crop()`'s actual forward reach.

This was found while investigating #138 (below), using an open room with the agent centred (no walls within reach) to isolate "visibility computation bug" from "legitimate off-map padding" - the two were previously confounded by only ever testing corner scenarios.

### #138: RGB first-person padding colour (folded in)
Out-of-view/unseen tiles used `dark_cell_colour=0` (solid black). First attempt: swap in MiniGrid's raw wall grey constant `(100, 100, 100)`. **That was still wrong** - caught via review (thanks for pushing on this) and fixed properly in a follow-up commit:

Every real cell in `rgb_first_person`'s `patchwork` goes through `apply_minigrid_opacity(opacity=0.7)` before the FOV mask is applied, but `dark_cell_colour` is inserted as a flat literal *after* that, bypassing it entirely. So a genuinely-visible, opacity-adjusted wall doesn't render at the raw `100` at all - it renders at `~146`. Using `100` for the unseen fill created a visible seam between real walls (146) and unseen padding (100) within the *same image* - not a MiniGrid-vs-Navix gap, an internal Navix inconsistency.

Root-caused by checking both sides directly:
- Read `minigrid.core.grid.Grid.render_tile`: `MiniGridEnv.get_pov_render` passes `highlight_mask=vis_mask`, so every visible cell gets `minigrid.utils.rendering.highlight_img` applied (a 30% blend toward white). `apply_minigrid_opacity`'s formula (`255 - 0.7*(255-x)`) is algebraically identical to `highlight_img`'s (`0.7x + 76.5`) - not a coincidence, the existing docstring already said this function exists to replicate MiniGrid's opacity convention.
- Extracted the actual `view_cone` visibility mask and sampled every cell of a real rendered image individually, to confirm a *genuinely visible* wall cell (not an off-map padding cell that an earlier hand-check had mistakenly sampled instead) renders at `(146, 146, 146)` - matching MiniGrid's highlighted wall (`146.5`) almost exactly once the correct cell is compared.

Fixed by deriving `dark_cell_colour` from `apply_minigrid_opacity` itself, rather than hardcoding a second, easily-out-of-sync magic number.

### Bonus: floor tile grid-line colour (not its own issue, found while comparing renders)
`draw_grid_lines()` (`navix/grid.py`) exists specifically to replicate MiniGrid's grid lines, but was dead code - only referenced in a commented-out line in `rgb_first_person` - and was actually broken: `line_thickness = jnp.ceil(TILE_SIZE * 0.031)` returns a float-dtype array, which can't be used as a slice bound (confirmed by calling it directly - `TypeError: Only integer scalar arrays can be converted to a scalar index`). Fixed with plain `math.ceil()` instead (`TILE_SIZE` is a static Python int, never needed to be a JAX op).

The commented-out call site was also the wrong place: applying it to `rgb_first_person`'s per-step entity sprites (player, keys, balls, doors) would draw a grey line across their artwork - MiniGrid never does this, grid lines are background, objects render on top and cover them. Wired in at the correct place instead: `rendering/cache.py`'s `render_background()`, applied only to the floor tile, once, when the static background cache is built - not to walls (MiniGrid's `Wall.render()` fills its whole tile opaquely, hiding any line underneath, and Navix's wall sprite is already a flat uniform fill, confirmed by direct pixel inspection, not assumed).

`luminosity`: the function's default (`100`, MiniGrid's raw grey constant) is correct for walls but wrong here - MiniGrid's grid line is a sub-pixel-width, anti-aliased strip, only *partially* covered by that raw colour at typical tile sizes, while `draw_grid_lines` does a hard, fully-opaque fill with no antialiasing. Tried the "principled" default first - made the match measurably *worse* (corner mean diff 1.40 -> 6.02). Solved `0.7*L + 76.5 = 99` for `L`, then verified empirically rather than trusting the algebra alone: `luminosity=32` gives corner mean diff 1.40 -> **0.34**, open room 2.42 -> **0.54** - the best result across all three fixes.

## Verification

Measured on a real GPU (`Navix-Empty-16x16-v0` vs `MiniGrid-Empty-16x16-v0`, radius=3, identical agent position/direction), across all three commits:

| Scenario | pre-fix | #147 + raw grey | + opacity-adjusted grey | + tuned grid-line colour | + corner fix (final) |
|---|---|---|---|---|---|
| Open room, no walls in reach | 36.4 | 2.42 | 2.42 | 0.54 | **0.293** |
| Corner scenario | much higher | 14.54 | 1.40 | 0.34 | **0.199** |

### Fixed the corner-intersection residual too (caught by inspecting the actual diff array, not just mean/max)
The "known, accepted residual" above turned out to be fixable. MiniGrid draws its top and left grid-line strips as two independent, anti-aliased `fill_coords()` calls - the corner pixel gets covered by *both*, ending up brighter (`114`) than either strip alone (`99`). `draw_grid_lines()` applied one flat luminosity to the whole line, so it could match the edge or the corner, but not both.

Added an optional `corner_luminosity` parameter, filling just the `line_thickness x line_thickness` corner block separately. Solved `0.7*L + 76.5 = 114` for `L`, verified empirically: `corner_luminosity=54` at the floor-tile call site.

Sampled cells directly confirm the final state: real walls and unseen/padding cells render at exactly `(146, 146, 146)` in Navix, matching MiniGrid pixel-for-pixel; floor tiles' grid-line pixels (edges *and* corners) now match within 1 everywhere except the agent's own tile, where its sprite - a separate, independently-authored asset with its own baked-in border, deliberately not touched by this fix - replaces the floor tile entirely.

Confirmed `tests/test_issues.py::test_147` actually catches the #147 bug (not vacuously passing): temporarily reverted the `RADIUS * 2` fix and re-ran it - failed as expected (mean diff 36.4 vs. the test's threshold of 20) - then restored the fix and re-verified it passes.

`test_147` compares a real `rgb_first_person` render against a real MiniGrid render (`minigrid` is already a `requirements_test.txt` dependency) rather than hand-computing `crop()`'s pad/roll/rotate/slice geometry by hand.

### `view_cone` int32 overflow at large radius (found by @Near32's review, fixed here)

[@Near32](https://github.com/Near32) independently confirmed the `RADIUS * 2` fix against MiniGrid's own `Grid.process_vis()` (exact match), and separately found that `view_cone`'s `fin_diff` accumulates path *counts*, not just reachability - each step sums a 3x3 neighbourhood, so the total grows ~9x per step, and it overflows `int32` (~2.1e9) at cone radius 12. Past that, a wrapped-negative cell reads identically to a genuinely unreachable one, so overflow silently *removes* visibility instead of erroring. Since this PR passes `RADIUS * 2` to `view_cone`, it halves the safe ceiling on the public `RADIUS` parameter (`RADIUS <= 11` before this PR, `<= 5` after) - doesn't affect the default `RADIUS=3`, but a real regression at settings someone might reasonably use.

Fixed with the proposed clamp: `jnp.minimum(array, 1)` after every step, turning the accumulator into a real boolean flood (immune to overflow at any radius, behaviour-preserving below the old threshold since only `> 0` is ever read).

Independently re-verified all of it on a real GPU rather than taking the numbers on faith: reproduced the exact overflow figures from the standalone repro script, confirmed the fix restores exact correctness (`(2*radius+1)^2` visible cells, the correct closed form for an open room) at every radius from 10 to 24 - including radius 12+ where it was previously undercounting - confirmed identical masks to an unclamped reference at radius 3-7 (behaviour-preserving), and benchmarked no performance regression.

One tangent this raised: whether a `jax.lax.conv`-based 3x3 kernel would be faster than the existing roll-based separable sum. Benchmarked both directly - roll is consistently ~1.3-1.4x faster across grid sizes 15x15 to 63x63, so the existing approach was already the right call; not changed.

`tests/test_grid.py::test_view_cone_does_not_overflow_at_large_radius` - verified it actually catches the bug (temporarily removed the clamp, confirmed it fails with the exact expected undercounted value, restored the fix, confirmed it passes).

Full test suite (49 passed, including the new test) verified on a real GPU.

## Test plan
- [ ] CI passes
- `tests/test_issues.py::test_147` - real Navix-vs-MiniGrid render comparison, open room
- `tests/test_grid.py::test_view_cone_does_not_overflow_at_large_radius` - overflow regression test
- Full suite (49 passed) verified on real GPU

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -101,8 +101,11 @@ diff, no step-by-step narration of what you're checking, no preamble:
 1. **Correctness** — Does the change actually fix what it claims to fix?
    Cover the PR's stated issue, and separately, any additional out-of-scope
    fixes it bundles in (e.g. "fixes issue #X" plus an unrelated drive-by
-   fix) — each judged against its own claim, not the PR's headline. Say
-   plainly if something doesn't actually fix what it claims to.
+   fix) — each judged against its own claim, not the PR's headline. Give
+   each claim its own bullet point, prefixed with ✅ if it holds up or ❌ if
+   it doesn't, so an incorrect claim is visible at a glance without reading
+   the prose. Say plainly if something doesn't actually fix what it claims
+   to.
 2. **Requested changes** — Concrete things that should change before merge:
    real bugs, missing test coverage, a wrong approach. Skip nitpicks and
    style preferences that aren't load-bearing. Write "None." if there
--- a/navix/grid.py
+++ b/navix/grid.py
@@ -456,12 +456,13 @@ def draw_grid_lines(
     (e.g. `~32`, empirically matched against a real MiniGrid render -
     see `rendering/cache.py::render_background`) to compensate.
 
-    `corner_luminosity`, if given, fills just the `line_thickness` x
-    `line_thickness` corner block with a separate value. In MiniGrid,
-    the top and left line strips are drawn independently and both get
-    anti-aliased, so the corner - covered by both - ends up brighter
-    than either strip alone; a single flat `luminosity` for the whole
-    line can't reproduce that without also being tuned to a
+    `corner_luminosity` fills just the `line_thickness` x
+    `line_thickness` corner block with a separate value, defaulting to
+    `luminosity` (i.e. no distinct corner treatment) when omitted. In
+    MiniGrid, the top and left line strips are drawn independently and
+    both get anti-aliased, so the corner - covered by both - ends up
+    brighter than either strip alone; a single flat `luminosity` for
+    the whole line can't reproduce that without also being tuned to a
     higher value at just the corner.
 
     Args:
@@ -478,10 +479,10 @@ def draw_grid_lines(
     # raises `TypeError: Only integer scalar arrays can be converted
     # to a scalar index`).
     line_thickness = math.ceil(TILE_SIZE * 0.031)
+    corner_luminosity = luminosity if corner_luminosity is None else corner_luminosity
     tile = tile.at[:line_thickness, :].set(luminosity)
     tile = tile.at[:, :line_thickness].set(luminosity)
-    if corner_luminosity is not None:
-        tile = tile.at[:line_thickness, :line_thickness].set(corner_luminosity)
+    tile = tile.at[:line_thickness, :line_thickness].set(corner_luminosity)
     return tile
 
 
--- a/navix/grid.py
+++ b/navix/grid.py
@@ -20,6 +20,7 @@
 
 from __future__ import annotations
 from functools import partial
+import math
 
 
 from typing import Callable, Dict, List, Tuple
@@ -440,6 +441,17 @@ def apply_minigrid_opacity(image: Array, opacity: Array = jnp.asarray(0.7)) -> A
 def draw_grid_lines(tile: Array, luminosity: Array = jnp.asarray(100)) -> Array:
     """Draws grid lines on the given tile.
 
+    `luminosity` defaults to MiniGrid's own raw grey constant
+    (`COLORS["grey"]`), used as-is for e.g. walls. But MiniGrid's grid
+    lines specifically are drawn as a sub-pixel-width, anti-aliased
+    strip - at typical tile sizes only *partially* covered by that raw
+    colour, blending it with the background - whereas this function
+    does a hard, fully-opaque fill of `line_thickness` whole pixels
+    with no antialiasing. Using the raw `100` here renders visibly
+    bolder/brighter than MiniGrid's actual output; pass a lower value
+    (e.g. `~32`, empirically matched against a real MiniGrid render -
+    see `rendering/cache.py::render_background`) to compensate.
+
     Args:
         tile (Array): The input tile to which grid lines are drawn.
 
@@ -447,8 +459,13 @@ def draw_grid_lines(tile: Array, luminosity: Array = jnp.asarray(100)) -> Array:
         Array: The tile with drawn grid lines.
     """
     # Draw lines (top and left edges) at 3.1% of the tile size as per
-    # minigrid.core.Grid.render_tile
-    line_thickness = jnp.ceil(TILE_SIZE * 0.031)
+    # minigrid.core.Grid.render_tile. TILE_SIZE is a static Python int
+    # (not a runtime/traced value), so line_thickness must be a plain
+    # Python int too - jnp.ceil() returns a float-dtype array, which
+    # can't be used as a slice bound (`tile.at[:line_thickness, :]`
+    # raises `TypeError: Only integer scalar arrays can be converted
+    # to a scalar index`).
+    line_thickness = math.ceil(TILE_SIZE * 0.031)
     tile = tile.at[:line_thickness, :].set(luminosity)
     tile = tile.at[:, :line_thickness].set(luminosity)
     return tile
--- a/navix/grid.py
+++ b/navix/grid.py
@@ -438,7 +438,11 @@ def apply_minigrid_opacity(image: Array, opacity: Array = jnp.asarray(0.7)) -> A
     return jax.numpy.asarray(255 - opacity * (255 - image), dtype=jax.numpy.uint8)
 
 
-def draw_grid_lines(tile: Array, luminosity: Array = jnp.asarray(100)) -> Array:
+def draw_grid_lines(
+    tile: Array,
+    luminosity: Array = jnp.asarray(100),
+    corner_luminosity: Array | None = None,
+) -> Array:
     """Draws grid lines on the given tile.
 
     `luminosity` defaults to MiniGrid's own raw grey constant
@@ -451,6 +455,14 @@ def draw_grid_lines(tile: Array, luminosity: Array = jnp.asarray(100)) -> Array:
     bolder/brighter than MiniGrid's actual output; pass a lower value
     (e.g. `~32`, empirically matched against a real MiniGrid render -
     see `rendering/cache.py::render_background`) to compensate.
+
+    `corner_luminosity`, if given, fills just the `line_thickness` x
+    `line_thickness` corner block with a separate value. In MiniGrid,
+    the top and left line strips are drawn independently and both get
+    anti-aliased, so the corner - covered by both - ends up brighter
+    than either strip alone; a single flat `luminosity` for the whole
+    line can't reproduce that without also being tuned to a
+    higher value at just the corner.
 
     Args:
         tile (Array): The input tile to which grid lines are drawn.
@@ -468,6 +480,8 @@ def draw_grid_lines(tile: Array, luminosity: Array = jnp.asarray(100)) -> Array:
     line_thickness = math.ceil(TILE_SIZE * 0.031)
     tile = tile.at[:line_thickness, :].set(luminosity)
     tile = tile.at[:, :line_thickness].set(luminosity)
+    if corner_luminosity is not None:
+        tile = tile.at[:line_thickness, :line_thickness].set(corner_luminosity)
     return tile
 
 
--- a/navix/grid.py
+++ b/navix/grid.py
@@ -501,6 +501,20 @@ def view_cone(transparency_map: Array, origin: Array, radius: int) -> Array:
     def fin_diff(array, _):
         array = jnp.roll(array, -1, axis=0) + array + jnp.roll(array, +1, axis=0)
         array = jnp.roll(array, -1, axis=1) + array + jnp.roll(array, +1, axis=1)
+        # this accumulates path *counts*, not just reachability - each
+        # step lets every unit of mass split into up to 9 copies of
+        # itself (a 3x3 neighbourhood sum), so the total grows ~9x per
+        # step and only the downstream `view > 0` threshold is ever
+        # read. In an open area, that overflows int32 (~2.1e9) by
+        # cone radius 12 - beyond that, a wrapped-negative cell is
+        # indistinguishable from a genuinely unreachable one, so
+        # overflow silently *removes* visibility rather than erroring.
+        # Clamping to a boolean flood (min with 1) after every step is
+        # behaviour-preserving below the overflow threshold (verified:
+        # identical `> 0` masks at every radius that doesn't overflow)
+        # and immune to it above, since a boolean value can never
+        # overflow regardless of radius.
+        array = jnp.minimum(array, 1)
         return array * transparency_map, ()
 
     # initialise the field to all zeros, except at the source (agent's position)
--- a/navix/observations.py
+++ b/navix/observations.py
@@ -115,9 +115,13 @@ def categorical_first_person(state: State) -> Array:
     col = jnp.where(on_grid, col, W)
     transparency_map = transparency_map.at[row, col].set(transparent, mode="drop")
 
-    # apply view mask
+    # apply view mask. crop() places the agent at the *bottom* row of
+    # the 2*RADIUS+1 view, so the far row is 2*RADIUS cells forward of
+    # the agent, not RADIUS - view_cone's diffusion needs to reach that
+    # far, or the forward half of the view is permanently marked
+    # unseen regardless of whether real walls are there.
     player = state.get_player()
-    view = view_cone(transparency_map, player.position, RADIUS)
+    view = view_cone(transparency_map, player.position, RADIUS * 2)
 
     # get categorical representation
     tags = state.get_tags()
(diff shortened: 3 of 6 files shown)


Near32:
Glad you caught this, @epignatelli , as I have hit the same issue independently and I was about to open an issue for it.

### Confirming from the other side: 
Same mask at `RADIUS=3` (28 of 49 cells live), and at `RADIUS=1` it's starker still — the agent sees its own row and one cell ahead, 6 of 9, with the top row permanently blank. I also compared `radius=2*RADIUS` against MiniGrid's `Grid.process_vis()` on a constructed room and got an exact match (0 cells differing at `RADIUS=3`), so `RADIUS * 2` looks right.

### One caveat this fix runs into at larger `RADIUS`, which may be worth a follow-up: 
`fin_diff` accumulates *path counts* on an int32 field (each iteration sums a 3×3 neighbourhood), and those grow multiplicatively until they overflow. Because only `view > 0` is read, a wrapped-negative cell is indistinguishable from an unreachable one — so overflow silently reports **less** visibility rather than erroring.

Doubling the cone radius halves the `RADIUS` at which this bites. Measured in an **open room with no occluders**, where every cell should be visible, using this PR's `view_cone(..., RADIUS * 2)`:

```
RADIUS   view    cone=2R        field min      wrapped     visible
     5  11x11         10                0            0   121/121
     6  13x13         12   -1,774,327,240           32   152/169
     7  15x15         14   -2,042,464,474          180   145/225
```

`set_radius()` is public and `RADIUS=6` is a 13×13 view — an ordinary BabyAI-scale setting. Before this PR the ceiling was `RADIUS ≤ 11` (cone = `RADIUS`); after it, it's `RADIUS ≤ 5`. At the default `RADIUS=3` nothing changes, so this isn't a blocker.

Since the magnitudes are discarded by `> 0` anyway, a boolean flood is immune. Adding `array = jnp.minimum(array, 1)` before the transparency multiply in `fin_diff` is enough. I checked it's behaviour-preserving rather than assuming: in a room with an occluding wall, the resulting `> 0` masks are **identical** to the current implementation at `RADIUS` 3, 4 and 5 — where the summing version doesn't overflow — and at `RADIUS` 6 and 7 it wraps nothing where the current one wraps 24 and 121 cells.

<details>
<summary>Scripts used for the measurements above</summary>

`fin_diff` is inlined in each of these because it's a closure inside `view_cone`, and `view_cone` returns only the thresholded mask — the raw int32 field it accumulates isn't reachable through the public API. `crop` is imported from navix.

Run against [`main` @ 9202daa](https://github.com/epignatelli/navix/commit/9202daa98e8d242dcd628573c39aa1029fe7b749), i.e. *without* this PR applied — each script passes `2 * R` to the flood explicitly rather than relying on the call site, so it models this PR's behaviour on unmodified `main`.

**Produces the "independent of room size" claim** — the flood at radius 12 in walled rooms from 21×21 to 61×61, including sizes where the walls sit far outside its reach:

```python
import numpy as np, jax.numpy as jnp

def probe(N, radius):
    tm = np.ones((N, N), dtype=np.int32)
    tm[0,:] = tm[-1,:] = tm[:,0] = tm[:,-1] = 0
    tm = jnp.asarray(tm); c = N // 2
    a = jnp.zeros_like(tm).at[c, c].set(1)
    for _ in range(radius):
        a = jnp.roll(a,-1,0) + a + jnp.roll(a,+1,0)
        a = jnp.roll(a,-1,1) + a + jnp.roll(a,+1,1)
        a = a * tm
    a = np.asarray(a)
    reach = c - 1                      # cells from centre to the wall
    return reach, int(a.max()), int((a < 0).sum())

print(f"{'room':>7} {'reach':>6} {'radius':>7} {'surplus':>8} {'max':>14} {'negatives':>10}")
for N in (21, 31, 41, 61):
    reach, mx, neg = probe(N, 12)
    print(f"{N:>3}x{N:<3} {reach:>6} {12:>7} {12-reach:>8} {mx:>14} {neg:>10}")
for radius in (9, 10, 11, 12, 13):
    reach, mx, neg = probe(21, radius)
    print(f"{21:>3}x{21:<3} {reach:>6} {radius:>7} {radius-reach:>8} {mx:>14} {neg:>10}")
``` 

**Produces the RADIUS table** — this PR's `view_cone(..., RADIUS * 2)` in an empty room:

```python
import numpy as np, jax.numpy as jnp
from navix.grid import crop

def probe(R):
    """PR #148's behaviour: cone radius = 2*RADIUS, crop radius = RADIUS."""
    N = 4 * R + 9                      # room comfortably larger than the view
    tm = np.ones((N, N), dtype=np.int32) 
    tm[0,:] = tm[-1,:] = tm[:,0] = tm[:,-1] = 0
    tm = jnp.asarray(tm); c = N // 2
    a = jnp.zeros_like(tm).at[c, c].set(1)
    for _ in range(2 * R):
        a = jnp.roll(a,-1,0) + a + jnp.roll(a,+1,0)
        a = jnp.roll(a,-1,1) + a + jnp.roll(a,+1,1)
        a = a * tm
    f = np.asarray(a)
    vis = crop((a > 0).astype(jnp.int32), jnp.asarray([c,c]), jnp.asarray(0), R, padding_value=0)
    vis = np.asarray(vis); W = 2*R+1
    return f.min(), int((f < 0).sum()), int(vis.sum()), W*W
    
print(f"{'RADIUS':>7} {'view':>7} {'cone=2R':>8} {'field min':>16} {'wrapped':>8} {'visible':>12}")
for R in (3, 4, 5, 6, 7):
    mn, neg, v, tot = probe(R)
    print(f"{R:>7} {2*R+1:>3}x{2*R+1:<3} {2*R:>8} {mn:>16} {neg:>8} {str(v)+'/'+str(tot):>12}")
``` 

**Produces the clamp-parity check** — same room but with an occluding wall stub, comparing the summing flood against the clamped one:

```python
import numpy as np, jax.numpy as jnp

def flood(R, clamp, occluders=True):
    N = 4*R + 9
    tm = np.ones((N, N), dtype=np.int32)
    tm[0,:] = tm[-1,:] = tm[:,0] = tm[:,-1] = 0
    if occluders:                      # a wall stub, to test occlusion parity
        tm[N//2 - 2, N//2 - 3 : N//2 + 4] = 0 
    tm = jnp.asarray(tm); c = N // 2   
    a = jnp.zeros_like(tm).at[c, c].set(1)
    for _ in range(2 * R):
        a = jnp.roll(a,-1,0) + a + jnp.roll(a,+1,0)
        a = jnp.roll(a,-1,1) + a + jnp.roll(a,+1,1)
        if clamp:
            a = jnp.minimum(a, 1)
        a = a * tm
    return np.asarray(a > 0), np.asarray(a)
        
for R in (3,4,5,6,7):
    m_sum, f_sum = flood(R, clamp=False)
    m_cl,  f_cl  = flood(R, clamp=True)
    print(f"  RADIUS={R}: masks identical = {np.array_equal(m_sum, m_cl)}"
          f"   summing wrapped={int((f_sum<0).sum()):>4}  clamped wrapped={int((f_cl<0).sum())}"
          f"   visible sum={m_sum.sum():>5} clamp={m_cl.sum():>5}")
```       
          
</details>

Measured with the agent centred in a walled room. "field" is `fin_diff`'s int32 accumulator before the `> 0` threshold; "wrapped" counts cells that have gone negative, i.e. whose true path count exceeded `jnp.iinfo(jnp.int32).max` (2,147,483,647) and wrapped around — no genuine path count can be negative; "visible" is the 7×7 crop at `RADIUS=3` after thresholding and the frontier step.
 
```
cone radius   field max        field min      wrapped   visible
         10      80,156,209             0           0     49/49
         11     658,076,409             0           0     49/49
         12   2,089,261,746  -1,774,327,240        32     32/49
```      
         
(At radius 13 it degrades further: 77 wrapped cells.)

The cliff sits at radius 12 and is independent of room size — identical counts at 21×21, 31×31, 41×41 and 61×61, including sizes where the walls sit far outside the flood's reach. `2*RADIUS` is well below it for any sane `RADIUS`, so nothing to change in this PR; it would only matter if the radius later became a user-facing knob.

epignatelli:
Thanks @Near32 for catching the overflow as well! I'll include it in this PR so we can get going!

epignatelli:
Thanks for this - very thorough, and much appreciated. Fixed in the latest commit: clamped `fin_diff`'s accumulator with `jnp.minimum(array, 1)` exactly as you proposed.

Independently re-verified before merging (not just taking the numbers on faith): reproduced your exact overflow figures with your repro script, then confirmed the fix restores exact correctness at every radius from 10 to 24 (visible cell count matches the closed-form `(2*radius+1)^2` for an open room - e.g. radius 12 goes from 593 back to the correct 625), and confirmed identical masks to an unclamped reference at radius 3-7 with an occluding wall, matching your clamp-parity check.

Also benchmarked whether a `jax.lax.conv`-based 3x3 kernel would beat the roll-based separable sum, since your numbers made it clear how fast this grows - roll wins by ~1.3-1.4x consistently, so no change needed there.

Added `tests/test_grid.py::test_view_cone_does_not_overflow_at_large_radius` as a regression test. See the PR description for the full writeup.

epignatelli:
@claude can you have a look at this PR please?

PR #156, opened 2026-08-17
feat: implement PQN as a first-class navix agent

### LINKED ISSUE #155: feat: implement PQN as a first-class navix agent
Part of #134. Narrows the PQN half of #142 into an actionable task now
that the "port a first-class agent from research code" precedent is
set (`navix/agents/dreamer.py`, #152).

## What exists

`baselines/rejax/results*/` has committed PQN result JSONs for most
navix environments (e.g.
`navix_Navix-DoorKey-8x8-v0_pqn_5_0_1760737197.9737434.json`), produced
via the external `rejax` package. There is no `navix/agents/pqn.py` -
PQN has never been a first-class navix agent, on `main` or on the
`neurips` research branch (which does have `dreamer.py` but no
PQN/IQN agent module, only their baseline result data).

## Why this might matter now

Directly relevant to the leaderboard proposal (#130), and to #140
(reproducibility of `baselines/rejax` results) - right now the
committed PQN numbers can't be regenerated or extended (new envs, new
seeds) without going back to the external `rejax` package, unlike PPO
and Dreamer which navix now ships and can retrain itself.

## What PQN is

"Parallelised Q-Network" (Gallici et al., *Simplifying Deep Temporal
Difference Learning*) - an online, vectorized Q-learning agent with no
replay buffer and no target network, relying on LayerNorm and running
many parallel environments (purejaxrl-style) for stability. Whoever
picks this up should verify the exact recipe against the paper/rejax's
implementation rather than trusting this summary.

## Next steps

- Implement `navix/agents/pqn.py`, matching the `Agent`/`HParams` base
  class API (`navix/agents/agent.py`) that `PPO` and `Dreamer` already
  follow - `train`, `log_to_wandb`, shared logging conventions.
- Reuse `navix/agents/models.py` pieces where they fit (e.g. the
  encoder patterns already shared between PPO and Dreamer) instead of
  duplicating.
- Add a `tests/test_pqn.py` mirroring `tests/test_dreamer.py`'s
  coverage.
- Decide whether the existing `baselines/rejax/results*` PQN JSONs get
  replaced by a navix-native reproduction (see #140) or stay as
  external-baseline reference data alongside the new agent.

epignatelli:
## Summary
- Adds `navix/agents/pqn.py`: PQN ("Parallelised Q-Network", Gallici et al., [*Simplifying Deep Temporal Difference Learning*](https://arxiv.org/abs/2407.04811)) - an online, parallel-environment Q-learning agent with no replay buffer and no target network. Written from scratch against the paper + the reference single-file implementation ([CleanRL's `pqn.py`](https://github.com/vwxyzjn/cleanrl/blob/master/cleanrl/pqn.py)) as ground truth, **not** a wrapper around the external `rejax` package the existing `baselines/rejax/results*` PQN numbers came from.
- Adds `QNetwork` to `navix/agents/models.py`: Dense/LayerNorm/ReLU stacked twice + a linear Q-value head, matching the reference architecture. LayerNorm is load-bearing here, not incidental - it's the specific regularizer the paper shows keeps online Q-learning convergent without a replay buffer or target network.
- Exports `PQN`/`PQNHparams`/`QNetwork` from `navix.agents`.
- Adds `tests/test_pqn.py` (9 tests, mirroring `test_dreamer.py`'s style/coverage): `Agent` interface conformance, a full-training smoke test (no NaNs), the epsilon-greedy exploration schedule, `QNetwork`'s LayerNorm/no-output-activation, the collection-time value cache, the no-target-network bootstrap, and - the load-bearing numerical piece - a direct check that the Q(lambda) target (built from `rlax.lambda_returns`, using navix's own buffer indexing convention) matches the reference implementation's backward recursion exactly.

Closes #155.

## Verification
- Structurally mirrors `navix.agents.ppo` (rollout → flatten → minibatch SGD, same `TrainingState`/`Buffer`/`Agent` conventions) since the two share that training-loop shape, but the algorithm itself is a different family - **Q(lambda) targets are computed once per rollout, not re-evaluated every epoch** (unlike PPO's GAE, which needs re-evaluation for its importance-ratio correction; PQN has none).
- Found and fixed a real bug during verification: `QNetwork` initially expected pre-flattened observations like PPO's `ActorCritic` does (which requires the caller to wrap the env with `FlattenObsWrapper`) - navix's default observations are `(H, W[, C])` grids, so this silently broadcast `nn.Dense` over the wrong axis instead of erroring cleanly. Fixed by flattening internally in `QNetwork.__call__`, matching `Dreamer`'s `_flatten_obs` ergonomics instead of PPO's caller-side wrapping requirement.
- Verified the Q(lambda) target formula against a hand-rolled port of the reference recursion in an isolated environment before wiring it into the agent (float32-precision match, ~1e-6, across `q_lambda` values and mid-rollout/final-step episode boundaries) - this is the test committed as `test_qlambda_target_matches_reference_backward_recursion`.
- Full test suite: 75 passed (66 existing + 9 new), no regressions.
- Real training run on `Navix-Empty-5x5-v0` at the agent's own default hyperparameters (`budget=1_000_000`, matching `PPOHparams`' convention): 72% last-20% success rate in ~23s wall time on a single GPU (488 updates) - notably PQN trains over 10x faster per frame than `Dreamer` on the same task, consistent with the paper's speed claims. Not yet tuned to match PPO's ~99%/Dreamer's ~85-100% ceiling on this task; that's a follow-up, not a correctness gap (verified via the Q(lambda) equivalence test above, and via a 100k-frame run scoring poorly for the mundane reason that PQN's own reference recipe uses ~1000 updates even for trivial CartPole, and 100k frames is only 48 updates here).

## Test plan
- [x] `pytest tests/test_pqn.py` - 9/9 passed
- [x] `pytest tests/` - 75/75 passed, no regressions
- [x] Real training run on Empty-5x5 at default hyperparameters - 72% success, no NaNs
- [ ] CI matrix passes on this PR

https://claude.ai/code/session_01Be5vnY6tpvoTAwWAFN5mtH

### DIFF
--- a/navix/agents/__init__.py
+++ b/navix/agents/__init__.py
@@ -19,7 +19,7 @@
 
 
 from .ppo import PPO, PPOHparams as PPOHparams
-from .models import MLPEncoder, ConvEncoder, ActorCritic
+from .models import MLPEncoder, ConvEncoder, ActorCritic, QNetwork
 from .dreamer import (
     Dreamer,
     DreamerHparams as DreamerHparams,
@@ -27,3 +27,4 @@ from .dreamer import (
     Actor as DreamerActor,
     Critic as DreamerCritic,
 )
+from .pqn import PQN, PQNHparams as PQNHparams
--- a/navix/agents/models.py
+++ b/navix/agents/models.py
@@ -1,11 +1,12 @@
 """Shared neural-network building blocks used across navix's agents.
 
-Two groups live here: PPO's encoder/actor-critic components
-(`MLPEncoder`, `ConvEncoder`, `ActorCritic`), and Dreamer's world-model
+Three groups live here: PPO's encoder/actor-critic components
+(`MLPEncoder`, `ConvEncoder`, `ActorCritic`), Dreamer's world-model
 components (categorical-latent utilities, the symexp-twohot head, and
 the RSSM's encoder/decoder/prior/posterior networks) - the reusable
-pieces `navix.agents.dreamer.WorldModel` wires together into an RSSM.
-See `navix.agents.dreamer`'s module docstring for the algorithm these
+pieces `navix.agents.dreamer.WorldModel` wires together into an RSSM -
+and PQN's normalized Q-network (`QNetwork`). See `navix.agents.dreamer`
+and `navix.agents.pqn`'s module docstrings for the algorithms these
 latter components implement."""
 
 from functools import partial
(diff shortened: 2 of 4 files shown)


epignatelli:
@claude review please

epignatelli:
## Update: verified against rejax and retuned defaults

Investigated a "very low success" report on the initial version (72% on Empty-5x5 at 1M frames) before assuming it was a correctness bug:

- **Read rejax's actual `algos/pqn.py`** and traced its Q(lambda) target computation by hand. It caches Q at the state a transition *lands in*, which is one step off from where the algorithm should cache it. Cross-checked against the official upstream it explicitly credits (`mttga/purejaxql/pqn_gymnax.py`, Gallici's own repo): the original caches Q at the state a transition *starts* from. This implementation already followed the official (correct) convention, not rejax's.
- **Confirmed the gap was hyperparameters, not the target formula**: dropping rejax's own per-environment-tuned config for `Navix-Empty-6x6-v0` (`num_epochs=8`, `num_minibatches=128`, `exploration_fraction=0.3`, `end_e=0.1`, `gamma=0.9`) into *this* implementation reproduced rejax's reported ~100% success.
- **Retuned `PQNHparams`' defaults** accordingly (`num_epochs` 4→8, `num_minibatches` 8→32, `exploration_fraction` 0.5→0.3, `end_e` 0.05→0.1), informed by that evidence but not copying rejax's per-env HPO-searched learning rate verbatim. Now **99.5-99.9% success across 3 seeds** on Empty-5x5 at the same 1M-frame budget - matching PPO's ceiling on this task.
- Also swapped in `rlax.batched_index` and `rlax.l2_loss` for the hand-rolled action-gather and MSE loss.
- Full suite re-verified twice after these changes: 75/75 passed, no regressions.

Full details in the module docstring (`navix/agents/pqn.py`).

Issue #158, opened 2026-08-17
perf: benchmark every environment in detail (compile time, throughput, ...) to find optimization targets


epignatelli:
Part of #134 and #144.

## What exists vs. what this asks for

`neurips`'s `experiments/` (see #144's inventory) has paper-oriented
*aggregate* benchmarks: Navix-vs-MiniGrid and CPU-vs-GPU throughput,
each collapsed across environments or reported for one representative
env at a time (`experiments/throughput/throughput_envs.md`,
`experiments/cpu_vs_gpu/`). None of it is a systematic, per-environment
breakdown of where compile time or steady-state throughput is good or
bad across the full registered environment set.

\#150 and #157 did exactly that, but by hand, for one environment
(`KeyCorridor`) that a grep happened to flag as suspicious - and found
a real, unexpected problem (#157: 5-10x slower compile than other
envs, cause still unknown) that a systematic sweep would have caught
directly instead of needing a specific lead to go looking.

## Ask

Once #144 lands on what (if anything) of `experiments/`/`benchmarks/`
gets ported to `main`, use that as the base for a benchmark that
covers **every registered environment**, not a representative sample:

- Compile time (`jax.jit(jax.vmap(env.reset))` and `.step`, same
  methodology as #149/#150's measurements, for consistency).
- Steady-state throughput (resets/s, steps/s) at a fixed batch size.
- Whatever else turns out to be cheap to capture and useful for
  spotting outliers (jaxpr/HLO size, memory, wall-clock for a fixed
  number of steps, ...) - open to whatever's actually informative once
  someone's in the code, not a fixed list up front.

Output should make outliers obvious at a glance (e.g. one table/plot
across all envs, sorted by compile time or throughput) - the point is
finding the next `KeyCorridor`-shaped surprise before it needs its own
grep-and-guess audit like #150 did, not re-deriving aggregate numbers
the paper already has.

## Next steps

- Confirm scope with #144 first - this depends on what benchmark
  infrastructure (if any) actually gets ported, rather than building
  parallel/duplicate tooling.
- Run it, publish the results (even just as a comment/artifact here,
  doesn't need to block on a docs page), and file follow-up issues per
  outlier found - same pattern as #150 → #157.

Issue #140, opened 2026-08-15
Port baselines/rejax/train.py so committed results are reproducible


epignatelli:
Part of #134.

## Problem

`main`'s `baselines/rejax/` currently contains only the *result* JSON files (e.g. `baselines/rejax/navix_Navix-DoorKey-8x8-v0_ppo_32_0_....json`) - there's no `train.py` or equivalent script that actually produced them. As committed, these results aren't reproducible.

## Where the script lives

`baselines/rejax/train.py` (241 lines) exists on the `speedup`/`neurips` branches (`neurips`'s tip is a descendant of `speedup`'s tip, so either works as the source) but was never ported to `main` when the result files were.

## Next steps

Port `train.py` as its own small PR. Directly relevant to the leaderboard proposal (#130), which requires a `.sh`/script + commit SHA for every submitted result to be reproducible - `baselines/rejax`'s existing results currently don't meet that bar.

epignatelli:
This data is stale, and the one on the neurips or speedup branch is the one that superseeds it. Closing as there is nothing to do here

Issue #130, opened 2026-08-15
NAVIX leaderboard and benchmarks


epignatelli:
## Idea

Start a public leaderboard for NAVIX: a standard set of environments, a standard set of algorithms, run under a defined protocol, with results published somewhere people can compare against when proposing new methods (similar in spirit to what MiniGrid/Minigrid-BabyAI, Brax, or Craftax's leaderboards do, but leveraging NAVIX's speed to make full sweeps cheap to reproduce).

## What exists already that this could build on

- `navix/agents/` already has a `PPO` implementation (`PPO`, `PPOHparams`, `ActorCritic`) and the `Agent`/`HParams` base classes are designed for more algorithms to be added.
- `navix/experiment.py`'s `Experiment` class already supports multi-seed runs (`Experiment.run()`) and hyperparameter search (`Experiment.run_hparam_search()`).
- `baselines/ppo.py` is a working example that trains PPO across every registered environment.
- PR #99 added a `rejax` reproduction script (`baselines/rejax/`), with results already checked in for DQN/PPO/SAC across a range of environments (`baselines/rejax/results_navix/*.json`) - this could be a starting seed for the leaderboard's algorithm coverage rather than starting from zero.
- `navix.plotting` (PR #129) defines `MANDATORY_METRICS` (`perf/returns`, `perf/success_rate`, `perf/episode_length`, `iter/fps`, `iter/wall_time`) - the fixed, algorithm-agnostic set reasoned about to represent "how did training go", intended as the metric layer the per-algorithm detail page (see "Two distinct concepts" below) standardises on. Each is derivable purely from the (state, action, reward, done) interaction stream plus wall-clock time - the one interface every RL algorithm shares, regardless of internals - so the set stays meaningful for algorithms navix doesn't ship itself. Algorithm-specific ("diagnostic") metrics are deliberately *not* defined in navix core, since a submitted algorithm won't necessarily have any navix-specific code to declare them - that mapping belongs here, in the leaderboard, not in navix.

## Structure (decided)

- Submissions and results are collected in a `leaderboard/` folder at the repo root. This folder holds the submission's metadata (SHA, `.sh` script, results) - not necessarily the algorithm's implementation itself. As long as a submission provides a commit SHA and a `.sh` reproduction script, the algorithm's code can live wherever the submitter owns it (their own repo, a fork, a `rejax`-style external suite, ...) - navix doesn't need to vendor or take ownership of it.
- The results table has at least these columns:
  - **Algorithm name**
[...]

PR #159, opened 2026-08-17
Benchmark object for scoring algorithms across a preset environment set

epignatelli:
## Summary (WIP - not ready to merge)

Part of #158, and implements the "Benchmark" concept issue #130 (the navix leaderboard proposal) already scoped: *"a scenario-specific function that takes a trained agent and returns the metric(s) used to rank it."*

- `navix/benchmark.py`: `Benchmark` - an orchestration layer over the existing `Experiment` class (`navix/experiment.py`), running one `Experiment` per environment (with `budget` overridden per-environment) and scoring each with `perf/success_rate` (already bounded `[0,1]` and comparable across environments regardless of raw reward scale - answers #130's open cross-environment normalization question with no new machinery) averaged over the last 20% of training, aggregated into a single score.
- `AlgorithmEntry`: every algorithm a `Benchmark` scores is wrapped in this, carrying the provenance metadata #130's leaderboard-row spec requires (name, suite/author, commit SHA) plus implementation author and a paper link - so a `BenchmarkResult` comes out already shaped for a leaderboard row, not needing that metadata assembled separately later.
- Two starting presets, `NAVIX_1M` and `NAVIX_100K`, differing only in `budget`, both defaulting to every registered environment (`nx.registry()`).
- `tests/test_benchmark.py`: 6 tests covering `AlgorithmEntry`'s required metadata fields, `Benchmark.run`'s per-env `agent_factory` calls, budget override, result aggregation, and `default_score_fn`'s "last 20%" convention against hand-computed synthetic logs.

Per #130's existing "navix never vendors external algorithm code" decision (which superseded #140), an external algorithm like `rejax` fits into this design as a dependency - `AlgorithmEntry.agent_factory` wraps its public API against a navix `Environment`, no source copied in - rather than overriding that decision.

## Still to come (this PR is WIP)

- `benchmarks/` directory: `ppo.py`/`dreamer.py`/`pqn.py`, replacing `baselines/ppo.py`/`dreamer.py` (the only two *tracked* files under `baselines/` - not the untracked `baselines/rejax/`/`results_navix/` sometimes present in a local working tree, which were never part of the repo) with scripts that build an `AlgorithmEntry` and call `NAVIX_1M.run(...)`/`NAVIX_100K.run(...)`.
- Deleting the old `baselines/ppo.py`/`dreamer.py` once their replacements are verified equivalent.
- A `benchmarks/rejax_ppo.py`-style rejax-as-dependency adapter (own follow-up PR, not bundled here).

## Test plan
- [x] `pytest tests/test_benchmark.py` - 6/6 passed
- [ ] `pytest tests/` full suite (not yet re-run on this branch)
- [ ] `benchmarks/*.py` scripts + real training run
- [ ] CI matrix

https://claude.ai/code/session_01Be5vnY6tpvoTAwWAFN5mtH


epignatelli:
## Handover: design status and open work

Picking this up after a long design-iteration session. Here's where `navix/benchmark.py` stands as of `2096c84`, and what's genuinely still open.

### Current design

- **`Benchmark`** is abstract: `build_agent`/`train_on` are shared, concrete mechanics (build *some* agent for *some* environment, train it via `Experiment`); `run`/`summary`/`details` all `raise NotImplementedError` - no protocol-agnostic default exists for any of them, so a concrete protocol must override all three.
- **`FromScratchBenchmark(Benchmark)`** implements the one concrete protocol so far: a flat, order-independent list of environments, each trained from scratch with no transfer between them. `Navix1M`/`Navix100K` subclass it, fixing only `budget`.
  - `run(entry, env_ids=None, seeds=None) -> Dict[str, BenchmarkResult]` - one raw result per env_id.
  - `summary(raw) -> BenchmarkResult` - each field's `last_percent_mean` (last 20% by default), meaned across environments. This is the single score a leaderboard table row would show.
  - `details(raw) -> Dict[str, BenchmarkResult]` - identity for this protocol (`raw` is already the per-env curves); this is what a leaderboard would show on click-through for one entry.
- **`BenchmarkResult`** is a `flax.struct.PyTreeNode` (not a plain dataclass): `returns`/`episode_length`/`flops`/`memory_bytes`/`compile_time_seconds`/`fps`/`wall_time` are its leaves; `info: Dict` is marked `pytree_node=False` (static, ignored by `jax.tree.map`) - free-form, nothing populates it by default. A single `last_percent_mean(value, percent=20)` function reduces any `BenchmarkResult`'s curves via `jax.tree.map(last_percent_mean, result)`.
- **`AlgorithmEntry`** carries the provenance metadata issue #130's leaderboard-row spec requires (name, GitHub-validated author, full commit URLs for both navix and the algorithm's own repo, paper link) plus `agent_factory: Callable[[Environment], Agent]` - the seam that keeps navix from ever vendoring an external algorithm's code.
- `Benchmark.train_on(agent, env, seeds)` never logs to wandb - it's always the fast, local path (`Experiment.run(log_to_wandb=False)`). If wandb integration is wanted for a benchmark run, that's not wired up anywhere right now (see open item below).

Every design point above was verified by actually running it on a GPU machine (not just compiling), including a full `run.py` script end-to-end - see the commit messages on this branch for what was checked at each step.

### Open work

1. **`tests/test_benchmark.py` is currently broken** relative to the design above - it still references `.history()` (removed) and constructs `_TinyBenchmark(Benchmark)` (now abstract, won't work - needs `FromScratchBenchmark`). It was mid-rewrite when we paused to keep iterating on the design first, and is intentionally *not* included in the last commit. This needs a full pass before this PR is mergeable.

2. **The `Agent` interface doesn't document what `train()` must return.** `FromScratchBenchmark.score_env` assumes `logs` (from `Experiment.run`, ultimately from `Agent.train`) contains `done_mask`/`returns`/`lengths`/`iter/fps`/`iter/wall_time` - but `Agent.train`'s own signature/docstring (`navix/agents/agent.py`) says nothing about this; the only documentation of the *derived* metric names (`perf/returns`, `perf/episode_length`, etc.) lives in `navix/plotting.py`'s module docstring, several steps removed from the actual method a submitter implements. Right now this fails as a bare `KeyError` deep in `derive_scalar_metrics`'s caller if an external `Agent` doesn't follow the convention navix's own PPO/PQN/Dreamer happen to follow. Decided *not* to build runtime/compile-time validation for this (a `jax.eval_shape`-based pre-flight check was scoped but rejected) - the chosen direction is to lean on interface conformance instead ("implement the `Agent` interface correctly") - but that only works if the interface actually states the requirement, which it currently doesn't. Concretely: add a docstring to `Agent.train` naming the exact keys required, citing `plotting.MANDATORY_METRICS`/`Agent.log_to_wandb` as where the derivation happens.

3. **No wandb hook in the `Benchmark` path.** Before this round, `benchmarks/*/*/run.py` scripts called `.run(entry, log_to_wandb=True)` for live dashboards during benchmark scoring. That's gone now (`train_on` never streams). If live tracking during a benchmark submission run still matters, it needs a new, explicit hook - nothing today lets a caller get `logs` back for that purpose (only the reduced `BenchmarkResult` survives past `run()`).

4. **Only one protocol exists.** The module docstring describes what curriculum/continual-learning/open-ended-learning `Benchmark` subclasses *would* look like (different `run`/`summary`/`details` semantics, different result collection shapes - e.g. continual learning's natural shape is an R-matrix of task x task performance), but none of them are built. Not blocking for this PR, just context for anyone picking up the next protocol.

5. Once (1) is fixed, run the full test suite and a real end-to-end training run on GPU before merging - neither has happened against this exact commit yet (only targeted direct-script verification during iteration).

epignatelli:
@claude please do a thorough review

epignatelli:
_Posted by an AI agent (Claude Code) on behalf of @epignatelli, in response to the review above._

Went through each finding, verified it against the code before acting, and fixed what was real:

- ✅ **Missing `distrax`** (finding 1) — added to the root `requirements.txt` and all 10 `benchmarks/*/*/requirements.txt` (explicitly for `rejax-*` too, not relying on it arriving transitively via `rejax[compat] -> navix`). `0e05450`
- ✅ **README's wrong `seeds` default** (finding 3) — `10` → `16`, matching `Benchmark.seeds`' real `tuple(range(16))` default. `d3844c7`
- ✅ **`navix.plotting` breaking change, no shim** (finding 4) — restored as an alias to `navix.benchmarks.plotting`. Not a full re-export: `derive_scalar_metrics` moved *and* got renamed to `navix.agents.agent.derive_episodic_metrics`, so that one specifically stays gone — documented with a `BREAKING CHANGE:` footer. `94f04fc`
- ✅ **Dead `NON_CURVE_DIAGNOSTICS_KEYS` filter** (finding 5) — removed from `plot_diagnostics` (it filtered a dict that can never contain those keys); kept the constant itself (still correctly documents `diagnostics.npz`'s own shape) and the `ndim >= 1` guard (a real, reachable case). `aba9a01`

- ❌ **Placeholder/non-existent commit SHA** (finding 2) — false positive. `f12fbbaf72bbc42d966faaf23f7830734d4f21ac` is a real, resolvable commit in this repo's history (`f12fbba feat: restructure Benchmark into navix.benchmarks preset objects...`), confirmed both locally (`git log`) and via `gh api repos/epignatelli/navix/commits/...`. The review sandbox likely had `fetch-depth: 1` and genuinely couldn't see it — no fix needed.
- ⏭️ **`env_ids` label/row desync** (finding 6) — real but low-priority: only reachable if `run()`/`details()` are called separately with a changing registry in between, which doesn't happen for `Navix1M`/`Navix100K` or the `submit_entry` flow. Left as-is; noted for whoever adds a second protocol.

All changes verified with the relevant `tests/test_benchmark.py`/`test_benchmarks_plotting.py` subset (37-53 tests depending on scope) run against real JAX on GPU, not just static checks.

epignatelli:
@claude fixes implemented. Please do a new thorough review.

PR #162, opened 2026-08-28
fix(key_corridor): port MiniGrid's connect_all for door placement

### LINKED ISSUE #160: KeyCorridor stacks two doors on one cell, corrupting the rendered tile

epignatelli:
## Context

Prompted by an independent review of #161 (the KeyCorridor stacked-door fix): that PR's own code, and the "claude" GitHub Action review that approved it, only ever discussed two options for the buggy duplicate door - delete it (what #161 did) or move it to the goal-side room column. Neither is what MiniGrid's own `KeyCorridorEnv`/`RoomGrid.connect_all` actually does, and the review contained a factual error along the way (claiming non-goal goal-column doors are gated by a "permanently-unsatisfiable `requires=-1`" door - `requires=-1` is actually `EMPTY_POCKET_ID`, the *no lock needed* sentinel, not an unsatisfiable one).

Checking upstream MiniGrid directly (`minigrid/envs/keycorridor.py` + `minigrid/core/roomgrid.py`) shows navix's KeyCorridor never matched it structurally: MiniGrid places exactly *one* explicit door (the locked goal door) and gets every other connector - including the corridor↔key-room doors navix hardcodes one-per-row - from `RoomGrid.connect_all()`, a randomised search that keeps adding doors until every room is reachable from the agent's start, skipping anything that would touch the locked room. Navix's own version hardcoded a fixed, deterministic topology instead, with inter-row connectors only ever in the key-side column, never the goal-side one.

## What this does

Ports `connect_all`'s actual algorithm, adapted to be `jax.jit`-shaped:

- Candidate connector walls are enumerated statically - a fixed count per `n_rows` (2/6/10 for `n_rows` 1/2/3, i.e. every registered env ID).
- The one mandatory locked door (gating the goal room) and the always-open middle corridor are set up directly.
- Every other candidate is resolved via a single-pass, randomised Kruskal-style union-find: process candidates in random order, activate one iff it still joins two different components. Same reachability guarantee as MiniGrid's sample-with-replacement retry loop, but no unbounded iteration count, so it stays representable as a fixed-shape jitted program.
- Candidates that don't end up connecting anything still exist as `Door` entities (needed for a fixed per-`n_rows` entity count) but are permanently unopenable (a sentinel `requires` value no real key can ever match), so they behave as ordinary walls.

The union-find keeps `parent` fully flattened as an invariant instead of scanning up to `num_rooms` hops per lookup: a union starting from a flat array can only ever strand nodes one hop further off, so a single `parent[parent]` pass after each union always restores flatness. This was originally motivated by a GPU throughput regression I measured via wall-clock benchmarking (~8-25%) - profiling with `nsys` on an idle GPU afterwards showed the *actual* kernel-level cost of the original linear-scan union-find was statistically indistinguishable from this version's (157,953ns vs 157,732ns total device time, same kernels, same launch counts), meaning the wall-clock differences I was chasing were mostly measurement noise (warm-up/autotuning-cache contamination - whichever env ran first in a fresh process paid a one-time tax unrelated to the algorithm). Keeping the flattening optimisation anyway since it's strictly cheaper and the reasoning is sound regardless of how much of the original regression was real.

## Behavior change

Per `AGENTS.md`'s "flag behavior-changing PRs explicitly":

- Door counts, positions, and colours differ from `main` for every registered `Navix-KeyCorridorS*R*-v0` ID (different RNG draws, different topology).
- The previous asymmetry - inter-row connectors only ever in the key-side room column - is gone; either column can get one now, matching MiniGrid.
- This also fixes #160's underlying bug class more thoroughly than #161's targeted fix: every door position is unique *by construction* (distinct candidate walls, no reused draws), not patched for the one specific duplicate the old `for row in range(n_rows - 1)` loop produced.
- Compile time is up ~14-30% on a fresh trace (more candidates for XLA to trace/optimize) - a real, consistently measured cost across CPU and GPU, separate from runtime. Runtime throughput itself is not meaningfully affected (see profiling above).

## Testing

- Existing tests unmodified and passing: `test_98`, `test_141`, `test_160` (`tests/test_issues.py`), plus `test_actions.py`/`test_entities.py`/`test_environments.py` as a broader regression check.
- New direct verification (not yet added as a committed test - flagging for discussion): 500 seeds × all 6 registered env IDs, checking (a) every non-goal room reachable from the agent's start, (b) the goal room never gets a second, openable connector, (c) no two doors ever share a cell, (d) `jax.jit(jax.vmap(env.reset))` compiles and runs. All passed, 0 failures across 3000 episodes.
- Rendered before/after comparisons on `Navix-KeyCorridorS4R3-v0` confirming doors now appear on both room columns' inter-row walls, not just the key-side one.

https://claude.ai/code/session_017r9M6ZNasyVJjmBGCA585J

### DIFF
--- a/navix/environments/key_corridor.py
+++ b/navix/environments/key_corridor.py
@@ -50,6 +50,41 @@ def _room_id(row: int, col: int) -> int:
 
 
 class KeyCorridor(Environment):
+    """Find a key hidden behind unlocked doors, then use it to open the one
+    locked door guarding the goal.
+
+    Rooms form an `n_rows x 3` grid: the agent starts in the middle column
+    (always connected top-to-bottom - the "corridor"), keys live in the left
+    column, the goal lives in the right column behind the episode's one
+    locked door.
+
+    Every other connector - including corridor<->key-room doors - comes from
+    a `jax.jit`-shaped port of MiniGrid's `RoomGrid.connect_all`: a
+    randomised search that keeps adding doors until every non-goal room is
+    reachable from the agent's start, without ever adding a second connector
+    to the goal room. Door positions, counts, and which side of the grid
+    gets inter-row connectors therefore vary per episode, rather than
+    following a fixed layout.
+
+    Implemented as a single-pass, randomised Kruskal-style union-find
+    (activate a candidate wall iff it still joins two different components),
+    since MiniGrid's own sample-with-replacement retry loop has no static
+    iteration bound and can't be represented as a fixed-shape `jax.jit`
+    program. Candidate walls that end up not connecting anything still exist
+    as `Door` entities - a fixed count per `n_rows` is required for
+    `jax.jit` - but are permanently unopenable, so they behave as ordinary
+    walls.
+
+    Note:
+        Because those non-connecting candidates are still `Door` entities,
+        `observations.rgb`/`symbolic` render a closed, locked door sprite at
+        every candidate wall - including the ones that are functionally
+        solid walls. MiniGrid renders those as plain walls instead; there is
+        no way to tell them apart visually, only by checking
+        `state.entities["door"].requires` for the sentinel "no key can open
+        this" value.
+    """
+
     def _reset(self, key: Array, cache: Union[RenderingCache, None] = None) -> Timestep:
         n_rows_config = {3: 1, 5: 2}
         n_rows = n_rows_config.get(self.height, 3)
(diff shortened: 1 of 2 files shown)


epignatelli:
@claude review please

Issue #91, opened 2024-09-10
Difference from standard dynamic obstacles implementations


annasoligo:
Hi, Thanks for the implementation!

I've noticed the DynamicObstacles implementations differs from the standard Minigrid versions found in e.g. MiniGrid Farama.

In the Farama implementation an agent takes an action towards a ball and this causes a termination, but obstacles will not spawn on agents.
It seems in Navix, agents that take an action that would place them on a ball remain in the same place without registering a ball hit event, but terminations occur when balls move onto the agent. (The former arises from `record_walk_into` not registering ball hits, and the latter from `_can_spawn_there` recording a ball hit).

Was this an intentional difference?

epignatelli:
Hi @annasoligo, thank you very much for raising this!

Could you explain exactly the differences between the two, such that we can draw an action plan?
A reproducible example would be a terrific help, to make sure the bug remains fixed in future versions.

annasoligo:
Hi @epignatelli, 
I noticed this when visualising some evaluations and noticing that the agent would repeatedly take actions that would cause them to step onto the ball, but their grid poition remains the same and the episode does not terminate. I can't provide a particularly useful example as I use a custom observation wrapper and visualisation code, but I can identify where the two implementations differ in the code! 

In the Farama implementation if an agent takes an action to step onto an obstacle the episode is terminated (this is implemented in lines 161-167 of the [dynamic_obstacles.py](https://github.com/Farama-Foundation/Minigrid/blob/master/minigrid/envs/dynamicobstacles.py) file). In the Navix implementation in actions `record_walk_into` is called, but this only considers Goal, Wall and Lava collisions and does not record ball hits.

Wrt balls spawning on agents, the Farama implementation updates ball locations using the [place_obj](https://github.com/Farama-Foundation/Minigrid/blob/715394f7b635b103287fd83c8df3f0ecb86b2535/minigrid/minigrid_env.py#L313) function, which only places objects in empty grid positions (ie not on the agent). The Navix implementation (`update_balls` function) moves the balls and records ball hits when the new position is the agents current position.

At the moment I've fixed this locally by:
Updating record_walk_into in the EventsManager to record_ball_hit if the entity is a ball.
Changing update_balls and can_spawn there in the transitions file to only move the ball to empty squares, and to not override any ball hit events that just occurred from moving the agent. 
Its quite a hacked together fix at the moment but happy to clean it up and share.

annasoligo:
[...]

PR #170, opened 2026-08-31
Redesign EventsManager to support multiple simultaneous events per step

### LINKED ISSUE #139: Redesign events to support multiple simultaneous events per step
Part of #134.

## Problem

`EventsManager` currently stores one `Event` per event-type as a single scalar field (`goal_reached`, `ball_hit`, `wall_hit`, etc.), replaced wholesale each step. This is structurally why bugs like the ball-hit issue (#91, fixed ad-hoc in PR #125 - walking into a ball wasn't recorded because something else had already claimed/overwritten the relevant slot that step) happen: the design can't represent two different events occurring in the same step.

## Prior unfinished work

An unmerged branch, `refactor-events` (June 2024, WIP - old code left commented out mid-refactor, not finished), already explored the right direction: redesigning `EventsManager.events` as `Dict[Tuple[EntityType, EventType], Event]`, with each `Event`'s fields broadcast to match the relevant entity's batch shape (via a new `Event.empty_like(entity)` constructor), rather than one scalar `Event` per type. This lets multiple simultaneous events (different types, or the same type across different entities) coexist without clobbering each other.

## Next steps

This needs real completion (finish the refactor, remove the dead commented-out code, add tests for the multi-event case) and probably its own design discussion before implementation - not a drop-in port. Worth scoping out what the finished `EventsManager` API should look like (e.g. how callers query "did X happen this step" against a dict keyed by entity+event type) before starting.

epignatelli:
## Summary
Closes #139. `EventsManager` used to store one scalar `Event` per named field (`goal_reached`, `ball_hit`, `wall_hit`, ...), replaced wholesale on every `record_*` call - structurally unable to represent two independent occurrences of the same event type within one step:

- Two entity instances of the same type both triggering the same event (e.g. two balls both hitting the player) collapsed to one, since `transitions.update_balls` picked only the first ball whose `ball_hit` fired out of a vmapped population (`jnp.where(..., size=1)[0][0]`).
- An earlier `record_*` call in the same step's compound transition pipeline (e.g. the player walking into a ball) got silently discarded by a later one writing the same field (e.g. a different ball moving onto the player) - "something else had already claimed/overwritten the relevant slot that step", per issue #91/PR #125's original ad-hoc fix.

`EventsManager.events` is now `Dict[Tuple[str, str], Event]`, keyed by `(entity_type, event_type)` - each slot's `Event` is batched to match that entity type's own instance count (`Event.empty_like`), and `record_*` merges new hits into a slot (OR-ing `happened`, writing `position`/`colour` only where newly `True`) instead of replacing it wholesale (`EventsManager.merge_event`). Both bugs above are fixed by this.

`EventType`'s constants change from jax arrays to plain strings (dict keys must be static/hashable, not traced arrays), and `Event` drops its now-redundant `event_type` field (the dict key already encodes it - confirmed unused downstream via grep before removing). `EventsManager.happened(key)` is the new query surface, returning `False` (not `KeyError`) for an entity type a given environment never had. `State.__post_init__` populates `events` from `entities` automatically, so none of the ~10 call sites constructing `State(...)` directly needed to change.

Built on top of the direction an old unmerged WIP branch (`refactor-events`, June 2024) explored, but not a port of it - that branch had real bugs of its own (`jnp.array_equal` used where per-instance `positions_equal` was needed, in-place dict mutation instead of an immutable `.replace`-style return) and was based on a commit over a year stale relative to current `main`.

**Also fixed** (`f2ace3b`, raised by @claude's review as a "broader opportunity"): `EventsManager`'s own docstring says events are "which events happened this timestep", but nothing ever actually reset it between steps - once a slot's `happened` flipped `True` it stayed `True` for the rest of the episode. Never observable for the terminating conditions (`on_goal_reached`/`on_lava_fall`/`on_ball_hit` all end the episode the first time they fire), which is why PPO training stayed bit-identical to `main` across 5 environments despite this bug's presence - but a non-terminating reward like `rewards.wall_hit_cost` (currently unused by any registered environment, hence dormant) would have kept firing every subsequent step after the first wall hit. Fixed by resetting `state.events` to fresh right before `transitions_fn` runs each step in `Environment._step`, not after - a hit recorded during the step's own transition still counts, only carried-over history from earlier steps is dropped.

## Test plan
- [x] `pytest tests/` (excluding `test_dreamer.py`, pre-existing unrelated environment issue) on berlin (real JAX execution) - 114/114 passed on a clean `main`-based checkout of this branch
- [x] `tests/test_issues.py::test_139`, two scenarios against a synthetic seed-swept state: (1) two balls positioned so at least one PRNG seed (of 200 tried) moves both onto the player in the same step - both now record `happened=True`, not just one; (2) walking into ball 0 (`actions.forward`) followed by `transitions.update_balls` running afterwards in the same step's pipeline - ball 0's hit survives, not clobbered
- [x] `tests/test_events.py` (new): one hand-crafted, hand-verified-map policy per event type against a *real* registered environment/seed (not synthetic) - `Navix-Empty-5x5-v0` (GRID/HIT), `Navix-DoorKey-5x5-v0` (one coherent policy: WALL/HIT on the correct instance -> KEY/PICKUP -> DOOR/OPEN -> GOAL/REACH, checking every earlier event survives to the end), `Navix-LavaGapS5-v0` (LAVA/FALL), `Navix-Dynamic-Obstacles-5x5-v0` (BALL/HIT via walk-into, correct ball instance out of two), plus direct `record_door_unlock`/`record_ball_pickup` coverage (no current action reaches either) and `happened()`'s False-not-KeyError contract for entity types an environment never had
- [x] This surfaced a real, pre-existing bug (present in the old `record_key_pickup` too, just never caught - fixed in `8673f84`): `actions.pickup` called `record_key_pickup` with `keys` *after* it had already been reassigned to `DISCARD_PILE_COORDS`, so every recorded `KEY/PICKUP` event's position was the discard-pile sentinel `(0, -1)` instead of the key's real pickup location
- [x] `tests/test_events.py::test_events_reset_each_step_not_persisted_across_episode` (new): a real `env.step()` cycle confirms `WALL/HIT` fires on the step a wall is actually hit and resets to `False` on the very next step, instead of persisting for the rest of the episode
- [x] Re-ran the PPO regression check (5 environments, `budget=20_000` and `budget=500_000` with standard hyperparameters) against `main` after the events-reset fix too - still bit-identical, confirming no currently-shipped reward/termination composition is affected
- [x] `mypy`/`py_compile` on every touched file - no new errors vs. `main`'s established baseline

## Breaking change
- `EventsManager` no longer has `goal_reached`/`ball_hit`/`wall_hit`/`lava_fall`/`key_pickup`/`door_opening`/`door_unlock`/`ball_pickup` fields - use `events.happened((entity_type, event_type))` instead (see `navix.states.EventType`, `navix.entities.Entities`).
- `Event` no longer has an `event_type` field.
- `record_goal_reached`/`record_ball_hit`/`record_wall_hit`/`record_lava_fall`/`record_key_pickup`/`record_door_opening`/`record_door_unlock`/`record_ball_pickup` now take a per-instance boolean hit mask as their second argument instead of a position (`record_walk_into`/`record_pickup`/`record_grid_hit` keep their existing position-based signatures - they compute the hit mask internally, so `actions.py`'s `_can_walk_there` call sites didn't need to change).

### DIFF
--- a/navix/actions.py
+++ b/navix/actions.py
@@ -198,7 +198,7 @@ def pickup(state: State) -> State:
     # update events
     events = jax.lax.cond(
         jnp.any(key_found),
-        lambda: state.events.record_key_pickup(keys, position_in_front),
+        lambda: state.events.record_key_pickup(keys, key_found),
         lambda: state.events,
     )
 
@@ -298,7 +298,7 @@ def open(state: State) -> State:
     # update events
     events = jax.lax.cond(
         jnp.any(do_open),
-        lambda: state.events.record_door_opening(doors, position_in_front),
+        lambda: state.events.record_door_opening(doors, do_open),
         lambda: state.events,
     )
 
--- a/navix/actions.py
+++ b/navix/actions.py
@@ -185,6 +185,15 @@ def pickup(state: State) -> State:
 
     key_found = positions_equal(position_in_front, keys.position)
 
+    # update events - before keys is moved to the discard pile below, so
+    # the recorded event keeps the key's real pickup position, not
+    # DISCARD_PILE_COORDS.
+    events = jax.lax.cond(
+        jnp.any(key_found),
+        lambda: state.events.record_key_pickup(keys, key_found),
+        lambda: state.events,
+    )
+
     # update keys
     positions = jnp.where(key_found, DISCARD_PILE_COORDS, keys.position)
     keys = keys.replace(position=positions)
@@ -193,13 +202,6 @@ def pickup(state: State) -> State:
     key = jnp.sum(keys.id * key_found, dtype=jnp.int32)
     player = jax.lax.cond(
         jnp.any(key_found), lambda: player.replace(pocket=key), lambda: player
-    )
-
-    # update events
-    events = jax.lax.cond(
-        jnp.any(key_found),
-        lambda: state.events.record_key_pickup(keys, key_found),
-        lambda: state.events,
     )
 
     state = state.set_player(player)
--- a/navix/environments/environment.py
+++ b/navix/environments/environment.py
@@ -28,7 +28,7 @@ from flax import struct
 
 from .. import rewards, terminations, observations, transitions
 from ..rendering.cache import RenderingCache, TILE_SIZE
-from ..states import State
+from ..states import EventsManager, State
 from ..actions import DEFAULT_ACTION_SET
 from ..spaces import Space, Discrete, Continuous
 from ..entities import EntityIds
@@ -186,8 +186,27 @@ class Environment(struct.PyTreeNode):
         Returns:
             (Timestep): The timestep at time $t + 1$
         """
+        # events are a per-step record - EventsManager's own docstring
+        # says "which events happened this timestep" - but
+        # EventsManager.merge_event only ever ORs new hits onto
+        # whatever a slot already holds (by design, see issue #139:
+        # that's what lets two events in the *same* step's transition
+        # pipeline both survive), so nothing actually clears a slot
+        # back to False between steps unless something does it here.
+        # Reset before the transition runs, not after, so a hit
+        # recorded during this step's own transition_fn call still
+        # counts - only carried-over history from earlier steps is
+        # dropped. Without this, a non-terminating reward like
+        # rewards.wall_hit_cost would keep firing every subsequent step
+        # after the first wall hit, not just the step it happened
+        # (terminating conditions like on_goal_reached/on_lava_fall/
+        # on_ball_hit were never actually affected by this in practice -
+        # the episode ends the first time they fire, so there's no
+        # "subsequent step" for the stale True to leak into before
+        # autoreset gives every entity a fresh EventsManager anyway).
+        reset_state = timestep.state.replace(events=EventsManager())
         # update agents
-        state = self.transitions_fn(timestep.state, action, self.action_set)
+        state = self.transitions_fn(reset_state, action, self.action_set)
         t = timestep.t + 1
 
         # calculate termination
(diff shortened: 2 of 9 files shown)


epignatelli:
@claude review please

epignatelli:
Addressed the review's requested change: split the `key_pickup` position fix out of the `test(states):`-typed commit into its own `fix(actions):` commit (`8673f84`), so `CD.yml`'s commit-type-driven changelog categorizes it correctly. `tests/test_events.py` now lands as a clean `test(states):` commit (`6af431f`) on its own.

Also worth noting: CI caught a real, separate issue on this PR after the review ran - 3 of `test_events.py`'s tests hardcoded literal grid positions hand-verified on jax==0.4.30, which CI's jax==0.11.1 doesn't reproduce for the same seed (`jax.random.PRNGKey` is only deterministic *within* one jax version). Fixed by reading entity positions from the live reset state and deriving each policy dynamically instead - verified passing on CI's actual jax version before pushing.

Branch is now 3 commits, force-pushed after the split (rewritten history on this feature branch, not touching `main`).

epignatelli:
Fixed the events-reset issue you flagged under \"Broader opportunities\" too (\`f2ace3b\`): \`Environment._step\` now resets \`state.events\` to fresh right before running \`transitions_fn\` each step, so a slot's \`happened\` no longer persists for the rest of the episode once it fires once. Added a dedicated test (\`test_events_reset_each_step_not_persisted_across_episode\`) that goes through a real \`env.step()\` cycle to verify it directly.

Confirmed via grep that no current \`reward_fn\`/\`termination_fn\` reads \`prev_state.events\` (only the post-transition \`state.events\`), and re-ran the PPO bit-identical-vs-\`main\` regression check after this fix too - still bit-identical, since no currently-registered environment's reward composition is sensitive to cross-step event persistence (\`wall_hit_cost\`, the one that would be, isn't wired into any registration).

epignatelli:
@claude please review

PR #186, opened 2026-09-01
Implement Unlock and UnlockPickup

epignatelli:
## Summary
Closes #175, closes #176. Implements `Navix-Unlock-v0`/`Navix-UnlockPickup-v0` - the simplest case of MiniGrid's `RoomGrid` (`num_rows=1, num_cols=2`), built directly rather than a general `RoomGrid` port (#174 stays scoped to when `ObstructedMaze`/#183 actually needs the arbitrary case).

`Unlock`: matching key in the same room as a locked door, reward+termination on opening it. `UnlockPickup`: same layout plus a `Box` in the second room to pick up after unlocking. Full design rationale (shared layout builder, `Box` pickup infrastructure that didn't exist before this PR, the dtype bug it surfaced) is in the commit messages.

## Test plan
- [x] New `tests/test_unlock.py` - structural checks + full hand-crafted `env.step()` gameplay walkthrough across 10 seeds each, `EventsManager`/reward/termination wiring verified directly against real state transitions
- [x] Full test suite on berlin (excluding `test_dreamer.py`'s pre-existing unrelated environment issue) - in progress, will report/fix if anything regresses
- [x] `mypy`/`py_compile` on every touched file - no new errors vs. `main`'s established baseline
- [x] Real bug found and fixed along the way: `Box.create` never cast `colour` to `uint8` (unlike `Key`/`Door`/`Ball`) - `Box` was entirely unused/untested before this PR, so this surfaced immediately as a `jax.lax.cond` branch-dtype-mismatch the first time a real `Box` got constructed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01VJApsbeqR3TNKVhj5d6NbT

PR #189, opened 2026-09-01
fix(actions): drop() only moves the Pickable matching player.pocket

### LINKED ISSUE #188: actions.drop() can resurrect an already-consumed Pickable item when two Pickable entity types coexist
Found during independent review of PR #186 (`Unlock`/`UnlockPickup`).

`actions.drop()` (navix/actions.py) iterates over every `Pickable` entity type and moves *any* instance currently sitting at `DISCARD_PILE_COORDS` to the tile in front of the player:

```python
for k in state.entities:
    entity = state.entities[k]
    if isinstance(entity, Pickable):
        cond = jnp.logical_and(can_drop, entity.position == DISCARD_PILE_COORDS)
        position = jnp.where(cond, position_in_front, entity.position)
        entity = entity.replace(position=position)
        state.set_entity(k, entity)
```

It never checks `entity.id == player.pocket` - it just checks "is this entity at the discard pile", so it moves *every* already-consumed `Pickable` instance, not just the one actually in the player's pocket.

This was unreachable before PR #186 because no existing environment had two concurrently-`Pickable` entity types in the same episode. `UnlockPickup` is the first (`Key` + `Box`), which exposes it: reproduced via real `env.step()` calls on `Navix-UnlockPickup-v0` (`disable_autoreset=True`) - after key -> unlock -> box pickup (`player.pocket == box.id`), calling the `drop` action places **both** the box and the already-consumed key on the same tile, even though the key was never in the player's current pocket.

**Suggested fix**: match the drop candidate by `entity.id == player.pocket`, not just `position == DISCARD_PILE_COORDS`. Worth fixing in `actions.py` generally rather than per-environment, since it will resurface identically in any future multi-`Pickable`-type environment (e.g. the planned `ObstructedMaze`/full `RoomGrid` port in #174/#183).

epignatelli:
Found during independent review of PR #186. Closes #188.

`drop()` moved *every* `Pickable` instance sitting at `DISCARD_PILE_COORDS` to the tile in front of the player, rather than only the one matching `player.pocket` - unreachable before two `Pickable` entity types could coexist in one episode. PR #186's `UnlockPickup` (`Key` + `Box`) is the first, which exposes it: after key -> unlock -> box pickup, dropping resurrects both the box and the already-consumed key.

Also fixed while touching this function:
- `player.pocket` was never reset to `EMPTY_POCKET_ID` after a successful drop, leaving it stale.
- `_can_walk_there`'s returned events (e.g. a grid-hit record when the drop destination isn't walkable) were computed but never applied to state via `state.set_events`.

**Behavior change**: because `drop()` now applies `_can_walk_there`'s events, calling `drop` while facing a non-walkable tile/entity now records a grid-hit/walk-into event for that step, where previously it silently didn't. This only affects code paths reading `state.events` after a `drop` action (e.g. an opt-in `rewards.wall_hit_cost`, not any default reward/termination fn) - narrow blast radius, and arguably a bug fix rather than a break, so no `BREAKING CHANGE:` footer, but flagging per AGENTS.md's "flag behavior-changing PRs explicitly" guidance.

New coverage: `test_drop`, `test_drop_does_not_resurrect_a_different_pickable_instance` in `tests/test_actions.py` (this function had no test coverage before) - the regression test uses two `Key` instances rather than Key+Box, so it doesn't depend on PR #186's still-unmerged `Box` schema.

### DIFF
--- a/navix/actions.py
+++ b/navix/actions.py
@@ -215,7 +215,7 @@ def drop(state: State) -> State:
 
     Args:
         state (State): The current state.
-    
+
     Returns:
         State: The new state with the item in the pocket dropped in front of the player."""
     player = state.get_player(idx=0)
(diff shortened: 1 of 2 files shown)


epignatelli:
@claude please review

epignatelli:
Thanks for catching this - confirmed and fixed in 6172f18.

You were right: `test_drop_does_not_resurrect_a_different_pickable_type` depended on `Box` gaining an `id` field, which only happens in the still-unmerged PR #186. My own earlier verification passed only because I'd synced from a working tree that had PR #186's `entities.py` checked out, not this branch's actual content - a real gap in my process, not a false positive on your end.

Rewrote the regression test using two `Key` instances (id=1 already consumed/at the discard pile, id=2 in the player's pocket) instead of Key+Box - reproduces the exact same bug (matching by position alone, not id) without depending on unmerged work. Re-ran against this branch's actual checked-out content on berlin this time (not my working tree): all 9 tests in `tests/test_actions.py` pass.

Also updated the PR description to explicitly flag the events-application behavior change per your section 3 note (AGENTS.md's "flag behavior-changing PRs explicitly").

Issue #181, opened 2026-09-01
Implement BlockedUnlockPickup


epignatelli:
Part of #10. Depends on #174 (RoomGrid).

## Environment
- `MiniGrid-BlockedUnlockPickup-v0`

## Description
`UnlockPickup` (#176), plus the locked door is additionally obstructed by a `Ball` placed directly in front of it on the near side - the agent must move the ball out of the way (by picking it up and carrying it elsewhere, or dropping it aside) before it can even reach the door to unlock it.

## Effort: medium-high (once #174, #176 land)
Everything from `UnlockPickup` plus wiring up `Ball` pickup (same gap `Fetch`/#177 and `PutNear`/#178 need - four issues now share this one missing piece, worth doing once and reusing) and a "door approach is blocked by a non-walkable entity sitting on it" placement rule at generation time (`Ball.walkable` is already `False`, so this is really just "place a ball at the specific cell in front of the door" - the interesting part is confirming the agent's own movement/pickup/drop actions correctly let it clear the ball, not new blocking logic itself).

---
🤖 This issue was filed by Claude Code on the user's behalf.

epignatelli:
Implemented in #191 (merged, commit 85aaadf).

Issue #178, opened 2026-09-01
Implement PutNear family (6x6-N2, 8x8-N3)


epignatelli:
Part of #10.

## Environments
- `MiniGrid-PutNear-6x6-N2-v0`
- `MiniGrid-PutNear-8x8-N3-v0`

## Description
A room scattered with several objects; the agent must pick up one specific object and carry it adjacent to (drop it "near") a different specific target object, both named by the mission. Reward + termination on a correct drop.

## Effort: medium-high
The most interaction-heavy of the remaining families: needs `actions.pickup()` extended to `Ball` (same gap `Fetch`, #177, needs - worth sharing that work), `actions.drop()` (already exists and works today) extended with "near a specific target" success checking, and two independently-tracked mission targets (the object to carry, the object to drop near) rather than the single `state.mission` slot every other mission-based environment (`GoToDoor`, `GoToObject`/#173, `Fetch`/#177) needs. Likely needs either a second mission-like slot or encoding both targets into one structure - same design question `Memory`'s success/failure-position tracking raises, worth resolving consistently across both rather than inventing two different patterns.

---
🤖 This issue was filed by Claude Code on the user's behalf.

epignatelli:
Implemented in #191 (merged, commit 85aaadf).

Issue #177, opened 2026-09-01
Implement Fetch family (5x5-N2, 6x6-N2, 8x8-N3)


epignatelli:
Part of #10.

## Environments
- `MiniGrid-Fetch-5x5-N2-v0`
- `MiniGrid-Fetch-6x6-N2-v0`
- `MiniGrid-Fetch-8x8-N3-v0`

## Description
A single room scattered with `N` `Ball`/`Key` objects of varying type/colour; a mission string names one specific object ("fetch a red ball"). Reward + termination on picking up (not just reaching, unlike `GoToObject`) the matching object; picking up the *wrong* one is a no-op (episode continues).

## Effort: medium
Same target-selection shape as `GoToObject` (#173) - place several distinct objects, pick one as the per-episode target via `State.mission`-style tracking - but success requires *pickup*, not proximity. `actions.pickup()` today only handles `Key` (`Ball` pickup is dead code - `record_ball_pickup` exists but nothing calls it, confirmed via grep). This needs `Ball` wired into `pickup()`'s dispatch alongside `Key`, plus reward/termination checking `player.pocket` against the target object's id, matching the mission - the "wrong object picked up is a no-op, not a failure" semantics also needs the pickup action itself to only actually remove/pocket the object if the mission doesn't care which was picked (MiniGrid removes any picked-up object from the grid regardless of match, only the reward differs) - worth confirming this against MiniGrid's actual `fetch.py` behavior precisely when implementing, not assuming.

---
🤖 This issue was filed by Claude Code on the user's behalf.

epignatelli:
Implemented in #191 (merged, commit 85aaadf).

Issue #173, opened 2026-09-01
Implement GoToObject family (6x6-N2, 8x8-N2)


epignatelli:
Part of #10.

## Environments
- `MiniGrid-GoToObject-6x6-N2-v0`
- `MiniGrid-GoToObject-8x8-N2-v0`

## Description
A single room scattered with several `Ball`/`Key` objects of different colours; the agent must navigate to (not pick up) the one matching a mission string (e.g. "go to the red ball"), chosen randomly each episode. Reward on reaching (being adjacent to and facing, or standing next to) the target object; timeout otherwise.

## Effort: low-medium
Structurally close to the already-implemented `GoToDoor` (`navix/environments/go_to_door.py`), which already establishes the pattern this needs: multiple candidate objects placed in one room, one selected as the per-episode target via `State.mission: Event`, reward/termination checking proximity to `state.mission.position`/`state.mission.colour`. The main new work is placing several `Ball`/`Key` instances (both entity types already exist and are fully wired for placement/collision) of *distinct* colours in one room without overlap, and picking one as the target - no new entity types, actions, or cross-cutting infrastructure needed.

---
🤖 This issue was filed by Claude Code on the user's behalf.

epignatelli:
Implemented in #191 (merged, commit 85aaadf).

PR #191, opened 2026-09-02
Implement RedBlueDoors, GoToObject, Fetch, PutNear, BlockedUnlockPickup

### LINKED ISSUE #172: Implement RedBlueDoors family (6x6, 8x8)
Part of #10.

## Environments
- `MiniGrid-RedBlueDoors-6x6-v0`
- `MiniGrid-RedBlueDoors-8x8-v0`

## Description
Agent starts in a single room split by a wall into two zones, each with one exit `Door` (one red, one blue - unlocked, just closed). Must open the red door, then the blue door, in that order, for reward; opening blue before red ends the episode with 0 reward. Solvable without memory (order is checked against door state, not agent memory).

## Effort: low
Single room, two `Door` entities (reuses existing `Door`/`actions.open` entirely) - no new entity types or actions needed. The only new piece is order-tracking: reward/termination need to know "was the red door already open when the blue door just opened" - i.e. compare door-open state *before* vs *after* this step's transition, not just a single-state check like every other reward/termination predicate in `navix/rewards.py`/`terminations.py` today. Likely the cleanest new primitive here (and directly reusable by the `RedBlueDoors` case only, unlike `state.mission`) is a small `prev_state`-vs-`state` comparison in a dedicated reward/termination function pair, since both already receive `prev_state` as their first argument.

---

epignatelli:
Closes #172, #173, #177, #178, #181.

## Summary

Five new environment families, all single-room (no `RoomGrid`/#174 needed - `BlockedUnlockPickup` reuses `Unlock`/`UnlockPickup`'s existing `two_equal_rooms_with_door` instead), sharing new infrastructure:

- **`Ball` is now `Pickable`** (gained an `id` field, matching `Key`/`Box`) and wired into `actions.pickup()` - `record_ball_pickup`/its event slot already existed as dead code. `pickup()`'s shared closure now dispatches via the already-generic `state.events.record_pickup(entity, position)` instead of a manual per-type branch.
- **`EventsManager.happened_at(key, position)`**: like `happened()`, but identifies *which* instance fired (needed by Fetch/PutNear to tell a correct pickup from an incorrect one).
- **`State.mission2`**: a second, independently-tracked mission target (only `PutNear` needs two simultaneously - the object to carry vs. the object to drop near).
- **`grid.random_distinct_positions`**: a real bug fix - `random_positions(key, grid, n=n)` samples *with replacement*, so two objects could land on the same cell. Found via actual test failures, not inspection.
- Both this and a second bug (the default `stochastic_transition` unconditionally wanders every `Ball` entity a random step each turn, meant for `DynamicObstacles`) were caught only by running real gameplay on berlin - structural/jit-vmap smoke tests alone missed both.

## Per-environment notes

- **RedBlueDoors** (#172): verified against MiniGrid's actual `RedBlueDoorEnv` source (both doors sit in one dividing wall; order, not position, determines success) - not assumed.
- **GoToObject** (#173) and **Fetch** (#177): both had originally-guessed semantics in their own issue text that turned out wrong once checked against real MiniGrid source (`GoToObject` requires the `done` action, not just proximity, and `toggle` fails immediately; `Fetch`'s wrong-pickup ends the episode, it's not a no-op).
- **PutNear** (#178): the most involved - two independently-tracked targets, wrong-pickup failure, and a Chebyshev-distance drop-near success check, all verified against MiniGrid's actual `PutNearEnv.step`.
- **BlockedUnlockPickup** (#181): lives in `unlock.py` alongside `Unlock`/`UnlockPickup` (not its own file), reusing `two_equal_rooms_with_door` via a new `block_door: bool` parameter.

## Testing

Structural + real `env.step()` gameplay + jit/vmap tests per environment, all verified on berlin (real JAX/GPU execution). Gameplay tests use real BFS pathfinding for test navigation (not the row-then-column heuristics `test_unlock.py`'s existing tests use) - simpler and robust to any obstacle layout. Seed counts and per-test env-id coverage are deliberately trimmed (structural/jit-vmap tests already cover every registered size cheaply; gameplay tests use 2 seeds against one representative size) to keep full-suite CI time reasonable now that this adds five more environment families' worth of tests.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Run unit tests with pytest
         run: |
           wandb offline
-          pytest
+          pytest -n auto
       - name: Run examples
         run: |
           for example in examples/*.py; do
--- a/navix/actions.py
+++ b/navix/actions.py
@@ -170,32 +170,31 @@ def left(state: State) -> State:
 
 
 def pickup(state: State) -> State:
-    """Picks up an item (`Key` or `Box`) in front of the player and puts
-    it in the pocket.
+    """Picks up an item (`Key`, `Box`, or `Ball`) in front of the player
+    and puts it in the pocket.
     Args:
         state (State): The current state.
     Returns:
         State: The new state with the player entity having the item in the pocket."""
 
-    def pickup_entity(state: State, entity_enum: str, entity, setter) -> State:
-        """Shared logic for one pickable entity type (`Key`/`Box`): if the
-        player is facing an instance of `entity`, discard it and put its
-        `id` in the player's pocket. A closure (not a module-level
-        function) specifically so it stays out of `actions.py`'s public
-        surface, where every other name is a real `State -> State`
-        action - unlike those, this needs `entity_enum`/`entity`/
+    def pickup_entity(state: State, entity, setter) -> State:
+        """Shared logic for one pickable entity type (`Key`/`Box`/
+        `Ball`): if the player is facing an instance of `entity`,
+        discard it and put its `id` in the player's pocket. A closure
+        (not a module-level function) specifically so it stays out of
+        `actions.py`'s public surface, where every other name is a real
+        `State -> State` action - unlike those, this needs `entity`/
         `setter` too, to stay parametrized over which pickable entity
-        type `pickup`'s two call sites (`Key`, `Box`) are handling.
+        type `pickup`'s call sites are handling.
 
         Args:
             state (State): The current state.
-            entity_enum (str): The entity type's `Entities.*` string,
-                used to dispatch to the matching `EventsManager.
-                record_*_pickup`.
             entity: Every instance of this entity type in the
-                environment (`state.get_keys()`/`state.get_boxes()`).
+                environment (`state.get_keys()`/`state.get_boxes()`/
+                `state.get_balls()`).
             setter (Callable[[Any], State]): `state.set_keys`/
-                `state.set_boxes` - writes the updated entity batch back.
+                `state.set_boxes`/`state.set_balls` - writes the updated
+                entity batch back.
 
         Returns:
             State: The new state with the item picked up, if the player
@@ -207,15 +206,13 @@ def pickup(state: State) -> State:
 
         # update events - before entity is moved to the discard pile
         # below, so the recorded event keeps the item's real pickup
-        # position, not DISCARD_PILE_COORDS.
-        record = (
-            state.events.record_key_pickup
-            if entity_enum == Entities.KEY
-            else state.events.record_box_pickup
-        )
+        # position, not DISCARD_PILE_COORDS. record_pickup already
+        # dispatches on isinstance(entity, Key/Box/Ball) internally, so
+        # this stays correct for any Pickable type without needing a
+        # per-type branch here.
         events = jax.lax.cond(
             jnp.any(found),
-            lambda: record(entity, found),
+            lambda: state.events.record_pickup(entity, position_in_front),
             lambda: state.events,
         )
 
@@ -235,9 +232,11 @@ def pickup(state: State) -> State:
         return state
 
     if Entities.KEY in state.entities:
-        state = pickup_entity(state, Entities.KEY, state.get_keys(), state.set_keys)
+        state = pickup_entity(state, state.get_keys(), state.set_keys)
     if Entities.BOX in state.entities:
-        state = pickup_entity(state, Entities.BOX, state.get_boxes(), state.set_boxes)
+        state = pickup_entity(state, state.get_boxes(), state.set_boxes)
+    if Entities.BALL in state.entities:
+        state = pickup_entity(state, state.get_balls(), state.set_balls)
     return state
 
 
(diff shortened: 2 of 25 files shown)


epignatelli:
@claude please do a very thorough review

epignatelli:
Thanks - this caught a genuinely critical bug. Addressed all five findings, reproduced each one directly before fixing (not just from the review's description):

**1. `actions.pickup` N=2 broadcasting bug (critical)** - confirmed exactly as described: with 2 balls, picking up one corrupted both balls' rows to the discard row (columns untouched), via `Navix-Fetch-8x8-N3-v0` seed 0. Fixed with `found[:, None]` (726eb83), matching the pattern `EventsManager.merge_event`/`actions.drop()` already use correctly. Also hits the already-shipped `Navix-Dynamic-Obstacles-5x5(-Random)-v0` now that `Ball` is `Pickable` - new regression test uses two `Ball` instances specifically.

**2. `GoToObject`'s facing requirement** - installed `minigrid==3.1.0` myself and confirmed: real `GoToObjectEnv.step` only checks orthogonal adjacency, no facing check at all. Fixed `on_target_done` (0594f39) and added a regression test that explicitly faces away from the target before calling `done`.

**3. `RedBlueDoors`' geometry claim** - checked the real `_gen_grid` source directly; you're right, it's a 3-chamber layout with doors on two different walls, not one shared dividing wall. Confirmed this is genuinely behaviorally harmless (toggle only needs adjacent-and-facing, never walking through; termination fires the instant blue opens, before either outer chamber could ever be reached) - left the simpler implementation, but corrected the false "verified" claim in the docstring (6c2e4f6).

**4. `test_put_near.py` flaky drop-distance bug** - confirmed the mechanism (drop lands on `translate(player.position, player.direction)`, not `player.position`). Fixed (66903b2).

**5. Missing `BREAKING CHANGE:` footer** - added via a follow-up commit (9fa19c0), since amending the original commit isn't available to me in this environment and this repo merges with real merge commits (every commit gets parsed). Also fixed an unrelated inaccuracy I noticed in the same docstring while there: `Ball.probability` isn't actually used for movement at all - `update_balls` moves every ball unconditionally.

**Not fixed (deferred, matches your own "New risks" categorization, not "Requested changes")**: the object-placement-adjacency gap (`GoToObject`/`PutNear` can spawn objects already within Chebyshev-1 of each other, unlike MiniGrid's `reject_fn=near_obj`) - doesn't violate correctness, just occasionally makes an episode trivially easy. Happy to file a follow-up issue if you'd rather track it than leave it here.

All fixes verified with real gameplay on berlin (28/28 tests pass, including the two new regression tests) before pushing.

epignatelli:
Following up on my last comment - I'd deferred the object-placement-adjacency risk as non-blocking, but actually quantified it rather than just leaving it: **36.2%** of `Navix-PutNear-6x6-N2-v0` episodes (15.2% for 8x8-N3) spawned with the move object already within Chebyshev-1 of the target - trivially "solved" with no navigation needed. That's not a rare edge case, so I fixed it after all (fd69af6, f493f99): a new `grid.random_position_far_from` helper, used to re-place just the target object if the move/target pair spawns too close, keeping everything else untouched. New regression test asserts 0/500 seeds spawn already-solved for each size (verified on berlin).

All fixes from both rounds are now pushed and verified.

epignatelli:
One more from the review's "Broader opportunities" section - verified directly against source, and since we're implementing these fresh, implemented properly rather than just noted: real MiniGrid's `GoToObjectEnv`/`PutNearEnv` both sample `types = ["key", "ball", "box"]` **per object independently** (not `FetchEnv`, which only uses `["key", "ball"]` - correctly left unchanged).

Added `Box` as a third candidate type to both `GoToObject` and `PutNear` (d965b6a, 7aa2241, a4e51d5), matching MiniGrid's actual per-slot independent type sampling rather than a fixed split - implemented via the same padding-sentinel pattern `LavaCrossing` already uses (always allocate `n_objects` Key/Ball/Box slots, push whichever slots didn't get that type this episode to `DISCARD_PILE_COORDS`). Position alone still identifies the mission target regardless of type, so the reward/termination logic needed no changes beyond `events.on_any_target_pickup`/`on_put_near_wrong_pickup` also checking `Box` pickups (safe for `Fetch`, which shares `on_any_target_pickup` but never constructs `Box` entities - confirmed no regression).

All verified on berlin: 11/11 for `GoToObject`/`PutNear`, 4/4 for `Fetch` (no regression from the shared event-check change).

epignatelli:
Retired `mission2` entirely (per discussion) - `State.mission` is now `Tuple[Event, ...]` instead of a bare `Event | None` plus a numbered second field:

- Index 0 = primary/carry target, index 1 (where present) = second independently-tracked target.
- Empty tuple replaces the `None` sentinel for "no mission."
- A future environment needing a third simultaneous target is just a longer tuple, not a `mission3`.
- `GoToDoor`/`GoToObject`/`Fetch` construct `mission=(target,)`; `PutNear` constructs `mission=(move, target)`.

`GoToDoor` predates this branch (already on `main`) but needed the same update - verified directly on berlin (no dedicated test file exists for it) across all 3 registered sizes, real `env.step()` + jit/vmap, no regression. Everything else (`test_go_to_object.py`, `test_fetch.py`, `test_put_near.py`) reverified too: 15/15 passing.

epignatelli:
@claude please do a very thorough review

PR #196, opened 2026-09-02
feat(environments): implement Memory

### LINKED ISSUE #180: Implement Memory family (S17Random, S13Random, S13, S11, S9, S7)
Part of #10.

## Environments
- `MiniGrid-MemoryS17Random-v0`
- `MiniGrid-MemoryS13Random-v0`
- `MiniGrid-MemoryS13-v0`
- `MiniGrid-MemoryS11-v0`
- `MiniGrid-MemoryS9-v0`
- `MiniGrid-MemoryS7-v0`

## Description
A genuine memory test, structurally unlike anything else in navix: the agent starts in a small room seeing one object (`Key` or `Ball`, colour fixed to green), walks down a narrow hallway that ends in a T-split, and must go to whichever of the two end objects matches the one it saw at the start - the other one ends the episode with 0 reward. Directly relevant to #169 (transformer/history-encoding POMDP encoder) - this is the natural benchmark for evaluating whether a policy actually carries information across steps, as opposed to every other navix environment being solvable from the current observation alone (or nearly so).

Read MiniGrid's actual `memory.py` in full before implementing (`minigrid/envs/memory.py`, ~164 lines) - the grid layout (start room walls, hallway width/length, T-split geometry) and the success/failure position derivation (`success_pos`/`failure_pos` are the cells *adjacent to* each end object, not the objects' own cells, since `Key`/`Ball` aren't walkable) are both precise and worth matching exactly, not approximating.

## Effort: moderate-high, unique value
No existing navix environment shares this shape, so it's new procedural generation from scratch (though simpler than a general room-grid - a fixed T-shaped corridor). The main design decision: navix's `State.mission: Event` already carries one per-episode target position (used by `GoToDoor`) - Memory needs *two* target positions (success and failure), where reaching either terminates the episode but only one gives reward. Two candidate designs: (a) extend `State` with a second target slot, or (b) since success/failure positions are always mirror images of each other across the hallway's centre row (`failure_row = 2 * hallway_mid_row - success_row`, same column), store only `success_pos` in `state.mission` and derive `failure_pos` algebraically in the reward/termination function - avoiding any new `State` field. Worth deciding this once, since #178 (PutNear) hits a related "more than one mission target" question and ideally both resolve it the same way rather than inventing two different patterns.

---

epignatelli:
## Summary

Closes #180.

A genuine memory test, structurally unlike anything else in navix:
the agent starts in a small room seeing one green object (`Key` or
`Ball`, chosen at random each episode), walks down a hallway that
ends in a T-split, and must walk onto the cell adjacent to whichever
end object matches the one it saw at the start - the other one ends
the episode with 0 reward. Registers all 6 real MiniGrid
configurations: S17Random, S13Random, S13, S11, S9, S7
(`random_length` only for the two "Random" variants).

Grid geometry (start room walls, horizontal/vertical hallway,
T-junction) is built via vectorised boolean conditions over the whole
grid rather than python loops, since `hallway_end` is a per-episode
traced value for the random_length variants - translated directly
from MiniGrid's actual `memory.py::_gen_grid`, not approximated.

## Design notes

- `state.mission` stores only the success position - `failure_pos` is
  never a separate mission slot, since it's always the mirror image
  of the success position across the hallway's centre row (`height //
  2`), derived algebraically in the new `events.on_memory_failure`
  instead. Matches the same choice already made for PutNear's
  `move`/`target` pair, per the `mission`-tuple refactor from #191.
- `pickup` is remapped to `toggle` in this environment's `action_set`
  (both effectively no-ops here), matching MiniGrid's own `MemoryEnv.
  step`: `if action == Actions.pickup: action = Actions.toggle`.
- New `events.on_memory_success`/`on_memory_failure` are pure position
  equality against the player - no `done` action or facing needed,
  since the target is an ordinary walkable floor cell, not the
  object's own cell (unlike `GoToObject`'s `on_target_done`).
- `rewards.on_memory_success` is deliberately a flat 1/0, not
  MiniGrid's step-count-shaped `1 - 0.9 * (step/max_steps)` - matches
  navix's existing `on_goal_reached` convention, itself already the
  same simplification versus real MiniGrid.

## Testing

Structural + real `env.step()` gameplay (success and failure paths)
+ jit/vmap, covering all 6 registered sizes across 20 seeds each for
structural checks.

All verified via real JAX execution, though not exclusively through
the pytest harness: this branch was developed while chasing an
unrelated, already-tracked infra issue (see #194's description) -
`jaxlib` on the cluster used for real-execution checks has a
process-lifetime CPU-compiler flakiness (`Fatal Python error:
Aborted` inside XLA's compiler) that intermittently hits *any*
environment's `env.step()`, including pre-existing/already-merged
code (RedBlueDoors, PutNear, Unlock all hit it too during #194's
investigation) - not something specific to this PR. Where the full
pytest run for this file hit that flakiness, I independently verified
every piece of behaviour (structural correctness, the success path,
the failure path, jit/vmap compatibility, and random_length variance)
via small, standalone real-execution scripts instead, all of which
passed cleanly.

### DIFF
--- a/navix/environments/__init__.py
+++ b/navix/environments/__init__.py
@@ -36,3 +36,4 @@ from .red_blue_doors import RedBlueDoors
 from .go_to_object import GoToObject
 from .fetch import Fetch
 from .put_near import PutNear
+from .memory import Memory
--- a/navix/environments/memory.py
+++ b/navix/environments/memory.py
@@ -1,0 +1,247 @@
+# Copyright 2023 The Navix Authors.
+
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+
+#   http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+
+from __future__ import annotations
+from typing import Union
+
+import jax
+import jax.numpy as jnp
+from jax import Array
+from flax import struct
+
+from navix import observations
+
+from .. import actions, rewards, terminations, transitions
+from ..components import DISCARD_PILE_COORDS, EMPTY_POCKET_ID
+from ..entities import Ball, Directions, Entities, Key, Player, Wall
+from ..states import State, Event
+from ..grid import coordinates, room
+from ..rendering.cache import RenderingCache
+from .environment import Environment, Timestep
+from .registry import register_env
+
+
+# `pickup` is remapped to `toggle` (both a no-op here, since nothing in
+# this layout is `Toggleable` and no `Pickable` needs picking up) -
+# verified against MiniGrid's actual `MemoryEnv.step`: `if action ==
+# Actions.pickup: action = Actions.toggle`, done before dispatch.
+MEMORY_ACTION_SET = (
+    actions.rotate_ccw,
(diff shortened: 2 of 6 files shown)


epignatelli:
@claude please review

PR #198, opened 2026-09-02
feat(environments): add ObstructedMaze 1Dl and 1Dlh variants

epignatelli:
## Summary

Continues #183 (**1/13 → 3/13 registrations**), adding
`Navix-ObstructedMaze-1Dl-v0` and `Navix-ObstructedMaze-1Dlh-v0`.

MiniGrid registers all three `1D*` ids against the same class,
differing only by two kwargs - verified against its actual
registrations:

| env id | `key_in_box` | `blocked` |
|---|---|---|
| `1Dl-v0` | `False` | `False` |
| `1Dlh-v0` | `True` | `False` |
| `1Dlhb-v0` (already shipped, #197) | `True` | `True` |

So `ObstructedMaze1Dlhb` gains those two as static
(`pytree_node=False`) fields. They branch at trace-construction time,
so each id gets its own entity layout rather than padding unused
slots: no `Box` entity at all when `key_in_box=False` (the key lies
on the start room's floor instead), one `Ball` instead of two when
`blocked=False`. Same convention as `unlock.py`'s `block_door`.

## Testing

Structural and gameplay tests are now parametrised over all three
ids, with the walkthrough adapting to whichever obstacles a variant
actually has rather than being written out three times. Each
variant's full distinct solve path verified end to end via real JAX
execution:

- `1Dl` - take the key off the floor → unlock → cross → target
- `1Dlh` - open the box, take the revealed key → unlock → cross → target
- `1Dlhb` - clear the blocking ball first, then as `1Dlh`

`4 passed in 104s`, versus 102s for one variant at 2 seeds before -
`GAMEPLAY_SEEDS` drops 2 → 1 to pay for the extra variant coverage,
since each variant's *distinct solve path* is the coverage that
matters and extra seeds just replay the same path at real
`env.step()` cost. #196's CI job OOMed on exactly this shape (several
env ids × a gameplay walkthrough in one process), so this
deliberately doesn't triple the file's cost.

## Not in scope

The remaining ten registrations (`2Dl`, `2Dlh`, `2Dlhb`, `1Q`, `2Q`,
`Full`, each `-v0` and four also `-v1`) are all `ObstructedMaze_Full`-
based: a 3x3 room grid with up to 12 doors and multiple boxes/balls
live at once. That also needs the `open_box` caveat from #197's
review addressed first (its `jnp.sum`-over-`revealing` reveal math
assumes at most one hidden-key box opened per step, which holds for
every `1D*` variant but not for the multi-box ones). Tracked on #183.

### DIFF
--- a/navix/environments/obstructed_maze.py
+++ b/navix/environments/obstructed_maze.py
@@ -34,6 +34,7 @@ from typing import Union
 import jax
 import jax.numpy as jnp
 from jax import Array
+from flax import struct
 
 from navix import observations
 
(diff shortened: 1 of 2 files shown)


epignatelli:
@claude please do a very thorough review

PR #199, opened 2026-09-03
feat(environments): implement ObstructedMaze's Full-based variants

epignatelli:
## Summary

Stacked on #198 (targets that branch, not `main` - merge #198 first).
Continues #183 (**3/13 → 9/13 registrations**), adding the six
remaining `-v0` ids: `2Dl`, `2Dlh`, `2Dlhb`, `1Q`, `2Q`, `Full` - all
built on a new `ObstructedMazeFull` class (a 3x3 `room_grid`, up to 12
doors/8 boxes/9 balls).

Per this session's own convention (MiniGrid `-v1` behaviour ported
under a plain `-v0` navix name, difference documented), `2Dlhb`/`1Q`/
`2Q`/`Full` here carry MiniGrid's actual `-v1` semantics - so **all
13 real MiniGrid ids are now covered**, no separate `-v1`
registrations needed.

## Geometry

Verified against MiniGrid's actual `ObstructedMaze_Full`:
`room_size=6` throughout; the centre room connects to `num_quarters`
side rooms via *unlocked* doors, each side room connects on to 2
corners via *locked* ones (`2*num_quarters` locked doors total);
every locked door's key lives in the side room it opens from,
optionally hidden in a `Box`; blocking balls sit one step into the
side room from their own door.

MiniGrid's own `-v1` fix (the reason it exists at all): blocking-ball
positions are computed first, deterministically from the doors, and
excluded from every later placement - so a key can never end up
covered by a ball, which was possible in `-v0`.

## Testing

`tests/test_obstructed_maze_full.py`:
- structural checks (door/key/box/ball counts, `requires`<->`key.id`
  consistency, no two entities ever share a cell) across all 6
  registrations
- target-corner variety across seeds (`Full` only - the sole variant
  where all 4 corners are reachable)
- full real-gameplay walkthroughs for the three single-quarter ids
  (`2Dl`/`2Dlh`/`2Dlhb`) - each a genuinely different solve path (key
  off the floor / key in a box / clear a blocker first)
- direct N=2 and N=4 door/box isolation via state surgery for
  `1Q`/`2Q`/`Full` - real functional coverage for the thing that
  actually differs at this scale (does opening one of several live
  doors/boxes affect only that one?), without an expensive full BFS
  walkthrough through a 16x16 maze. This is exactly the class of bug
  #191 already found once (`actions.pickup`'s N=2 broadcasting bug) -
  worth checking directly, not just inferring from N=1 working
- jit/vmap for all 6

### A real bug, found and fixed during this work

`blocked = balls.position.shape[0] == 2`, copied from
`test_obstructed_maze.py`'s 1D convention (1 blocker + 1 target,
always 2 there), silently evaluated `False` for every single-quarter
variant with `blocked=True`, since two locked doors there means *two*
blockers + one target = 3 balls, never 2 - skipping blocker-clearing
entirely for `2Dlhb` every single time.

Worth being honest about how this was root-caused: a deterministic,
100%-reproducible `AssertionError` was initially chased as possible
jaxlib flakiness (the harness does have a real, separate, already-
tracked flakiness class from #194/#196), including several
standalone-script comparisons that misled rather than clarified,
before actually diffing the real function against a hand-typed
reproduction and finding the copy-paste bug. Fixed and confirmed via
4 clean isolated reruns of the specific failing test, plus individual
confirmation of every one of the file's 7 test functions.

The single gameplay-walkthrough function is split into 3 separate
top-level test functions (one per single-quarter variant) rather than
one function looping over all three - #196's CI OOM came from exactly
this shape (several real gameplay walkthroughs accumulating in one
process); separate functions at least give pytest-xdist the chance to
spread them across workers instead of concentrating the load.

## Not in scope

`open_box`'s reveal math was already generalised in this PR to handle
N>1 boxes correctly (see the isolation tests) - the caveat #197's
review raised about it only supporting N=1 is resolved here.

### DIFF
--- a/navix/environments/__init__.py
+++ b/navix/environments/__init__.py
@@ -37,4 +37,4 @@ from .go_to_object import GoToObject
 from .fetch import Fetch
 from .put_near import PutNear
 from .memory import Memory
-from .obstructed_maze import ObstructedMaze1Dlhb
+from .obstructed_maze import ObstructedMaze1Dlhb, ObstructedMazeFull
--- a/navix/environments/obstructed_maze.py
+++ b/navix/environments/obstructed_maze.py
@@ -257,8 +257,14 @@ class ObstructedMazeFull(Environment):
         covering it and making the episode unsolvable. navix implements
         only the fixed behaviour and registers it under the plain `-v0`
         name - blocking-ball cells are computed first (they follow
-        deterministically from the doors) and then excluded from every
-        subsequent player/key/box draw, so a key can never be covered.
+        deterministically from the doors) and then excluded, *along
+        with their own orthogonal neighbours*, from every subsequent
+        player/key/box draw. Excluding only a blocker's own cell (not
+        its neighbourhood) still let a key/box land on every remaining
+        side of a blocker sitting at a room-interior corner, fully
+        enclosing it - a real, ~1-4%-of-seeds bug on the multi-box
+        variants, found by review and fixed here (see
+        `test_obstructed_maze_full_blocking_balls_never_enclosed`).
         `2Dl`/`2Dlh` have no `-v1` upstream and port across unchanged.
 
     | navix id     | MiniGrid id     | `num_quarters` | `key_in_box` | `blocked` |
@@ -353,10 +359,30 @@ class ObstructedMazeFull(Environment):
             return jnp.where(room_mask(grid, ROOM_SIZE, *room), grid, -1)
 
         def keep_clear(room: Tuple[int, int]) -> Array:
+            # excludes each blocker's own cell *and* its 4 orthogonal
+            # neighbours, not just its own cell - a blocker only ever
+            # needs one of those neighbours free to be reachable at all
+            # (the others are already structurally non-walkable: one is
+            # its own door, closed until the blocker is cleared; any
+            # facing the room's outer wall are off the floor grid
+            # entirely), but excluding only its own cell left every
+            # neighbour open to a box - two boxes in the same room
+            # could (rarely) both land adjacent to the same blocker and
+            # fully enclose it, an unsolvable episode with no crash.
+            # Found by review on #199, quantified at ~1-4% of seeds on
+            # the variants with more than one box/blocker per room -
+            # verified absent on the already-shipped `Navix-
+            # ObstructedMaze-1Dlhb-v0` (0/1000 seeds), which only ever
+            # has one of each per room.
             cells = blockers.get(room, []) if self.blocked else []
             if not cells:
                 return DISCARD_PILE_COORDS[None]  # off-grid: excludes nothing
-            return jnp.stack(cells)
+            neighbours = [
+                cell + jnp.asarray(delta)
+                for cell in cells
+                for delta in ((0, 1), (0, -1), (1, 0), (-1, 0))
+            ]
+            return jnp.stack([*cells, *neighbours])
 
         # --- player ------------------------------------------------
         player_pos = random_positions(
--- a/navix/environments/obstructed_maze.py
+++ b/navix/environments/obstructed_maze.py
@@ -29,7 +29,7 @@ to follow-up work, once this one is confirmed to play correctly end
 to end."""
 
 from __future__ import annotations
-from typing import Union
+from typing import List, Tuple, Union
 
 import jax
 import jax.numpy as jnp
@@ -46,6 +46,7 @@ from ..grid import (
     open_wall,
     random_colour,
     random_directions,
+    random_distinct_positions,
     random_positions,
     room_grid,
     room_grid_dims,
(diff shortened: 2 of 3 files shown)


epignatelli:

## 1. Correctness

Verification method: installed the branch into a real arm64 Python 3.14 / JAX 0.11.1 environment (this machine's default toolchain is x86_64-under-Rosetta, whose jaxlib build fails on `AVX` checks under Rosetta translation — worked around with `uv venv --python 3.14`), ran the full `tests/test_obstructed_maze_full.py` suite for real, and independently pulled MiniGrid's actual `obstructedmaze.py`/`roomgrid.py` source to cross-check the geometry claims by hand rather than trusting the PR prose.

- ✅ Geometry matches real MiniGrid. Traced `SIDE_ROOMS`/`CORNER_ROOMS`/the `side = (quarter + offset) % 4` formula against MiniGrid's actual `ObstructedMaze_Full._gen_grid` (`side_rooms = [(2,1),(1,2),(0,1),(1,0)]`, `corners = [(2,0),(2,2),(0,2),(0,0)]`, `door_idx=(i+k)%4` for `k in [-1,1]`) — the claimed row/col transpose and the reused offset formula are exact matches, not just plausible-looking.
- ✅ jit/vmap for all 6 registrations: ran `test_obstructed_maze_full_jit_vmap_compatible` myself — passes.
- ✅ Full gameplay walkthroughs for `2Dl`/`2Dlh`/`2Dlhb`: ran all 3 tests myself — passes, including reward/termination.
- ✅ N=2/N=4 door and box isolation for `1Q`/`2Q`/`Full`: ran `test_obstructed_maze_full_door_and_box_isolation` myself — passes.
- ✅ "No two entities ever share a cell" structural check: ran `test_obstructed_maze_full_structure` myself — passes for all 6 ids × 3 seeds.
- ✅ The described `shape[0] == 2` copy-paste bug and its fix: confirmed `tests/test_obstructed_maze.py` (pre-existing, out of this diff) still legitimately uses `balls.position.shape[0] == 2` — correct there because the 1D convention is always exactly 1 or 2 balls — while the new file correctly uses `balls.position.shape[0] > 1` instead, matching the PR's own description of the fix.
- ✅ `open_box` N>1 generalization: the isolation test opens one of two boxes per room and confirms only the matching key is revealed and only that box is discarded — holds up.

## 2. Requested changes

**A real, reproducible unsolvable-episode bug, not caught by this PR's own test suite.** A blocking `Ball` can end up fully enclosed — all 4 orthogonal neighbours non-walkable — by the room's own two `Box`es (plus the closed locked door on one side and the outer wall on another), with no cell left from which the player can ever face and pick it up. Since the blocker sits on the *only* interior-adjacent cell of its own locked door (by construction — see `SIDE_DELTAS`), the player then also can never stand there to toggle that door open. This makes the affected seed's episode permanently unsolvable (it would silently truncate at `max_steps` with zero reward, never crash).

Verified directly against the actual game engine's own walkability rule (`actions._can_walk_there`: a cell is walkable iff `grid == 0` and no entity there has `walkable == False`), not an approximation:
- Hand-traced a concrete failing case — `Navix-ObstructedMaze-1Q-v0`, seed 2: locked door at `(5,14)`, its blocker at `(6,14)`, with a `Box` at `(6,13)`, a `Box` at `(7,14)`, the closed door itself at `(5,14)`, and the outer wall at `(6,15)` — all 4 neighbours of the blocker are non-walkable.
- Quantified across 300 fixed seeds (`PRNGKey(0..299)`) per env id (script + counts reproducible on request):
  - `Navix-ObstructedMaze-2Dlhb-v0`: 4/300
  - `Navix-ObstructedMaze-1Q-v0`: 5/300
  - `Navix-ObstructedMaze-2Q-v0`: 5/300
  - `Navix-ObstructedMaze-Full-v0`: 11/300
  - `2Dl`/`2Dlh` unaffected (no blocking ball there, `blocked=False`)
- Checked the already-merged `Navix-ObstructedMaze-1Dlhb-v0` the same way (1000 seeds): 0 trapped. So this isn't an old, accepted risk — it's newly introduced by this PR's `ObstructedMazeFull`, specifically because a side room here can hold *two* boxes and *two* blocking balls at once (the 1x2 `1Dlhb` room only ever has one of each), and a blocker landing on a room-interior corner cell has only 2 non-wall/non-door neighbours — both of which the room's 2 boxes can end up occupying.

Root cause: `keep_clear(room)` (and the equivalent exclusion in `1Dlhb`) only excludes a blocker's own cell from `random_distinct_positions`, not its neighbourhood — so nothing stops a box from landing directly adjacent to a blocker on every remaining open side. Given roughly 1-4% of seeds are affected on the variants that have blocking balls at all, this should be fixed before merge (e.g. exclude each blocker's orthogonal neighbours, not just its own cell, when drawing box/key positions), since a benchmark env that's silently unsolvable a non-trivial fraction of the time is a genuine correctness problem for anyone training or evaluating against it.

## 3. New risks

The same failure mode (an object enclosed with zero reachable orthogonal neighbours) is a structural risk for any future room-grid-based obstacle placement that reuses this "exclude only the exact occupied cell" convention — see "Broader opportunities" below. It's also a *silent* failure: the episode truncates normally rather than erroring, so it won't surface as a crash in CI or in a user's training run, only as an unexplained fraction of zero-reward episodes.

## 4. Breaking changes

None. This PR is purely additive (one new `Environment` subclass plus 6 new registrations); no existing environment's observations, dynamics, or rewards change under any existing env id, and no public function signature changes. No `BREAKING CHANGE:` footer is present, consistent with there being nothing breaking here.

## 5. Broader opportunities

- The unreachable-object failure class found above generalizes beyond this PR: any current or future room-grid obstacle placement (this file's own header flags `MultiRoom`/#182 as an upcoming reuser of the same `room_grid`/`room_mask` primitives) that excludes only an obstacle's own cell, not its neighbourhood, is exposed to the same risk. Worth a shared, reusable "no entity may become fully enclosed" structural check (a per-reset flood-fill/BFS reachability assertion) rather than relying on per-environment gameplay walkthroughs to catch it — this PR's own testing narrative explicitly avoided full walkthroughs for the larger variants (`1Q`/`2Q`/`Full`) for CI-cost reasons, which is exactly the situation where a cheap structural reachability check would have caught this without needing a real solve.
- None beyond that.

epignatelli:
Fixed - real bug, good catch. Thanks for specifically probing for a sibling of the acknowledged N-vs-1 bug rather than stopping at the tests I'd already written.

## Root cause, confirmed

`keep_clear(room)` excluded only a blocker's own cell from box/key placement, not its neighbourhood. A blocker sitting at a room-interior corner already has 2 of its 4 orthogonal neighbours non-walkable by construction (its own closed door, the room's outer wall) - excluding only its own cell left the other 2 open to the room's boxes, and with 2 boxes per side room, both could land there.

## Fix (d80d9ca)

`keep_clear` now excludes each blocker's 4 orthogonal neighbours too, not just its own cell. Re-ran your exact reachability check at higher sample counts:

| env id | before | after |
|---|---|---|
| `2Dlhb` | 4/300 | 0/300 |
| `1Q` | 5/300 | 0/300 |
| `2Q` | 5/300 | 0/300 |
| `Full` | 11/300 | **0/1000** |

## Regression test added

`test_obstructed_maze_full_blocking_balls_never_enclosed` - exactly the cheap, reset-only reachability check your "Broader opportunities" section suggested, 300 seeds x the 4 affected variants. Deliberately more than this repo's usual 3-seed test budget, since a ~1-4%-of-seeds bug needs a real sample to reliably catch.

Re-ran every other test function individually afterward (structure, corner-variety, isolation, all 3 gameplay walkthroughs, jit/vmap) - no regressions from the tighter exclusion set.

## On the generalization point

Agreed this class of risk isn't specific to this PR - noted for whoever picks up `MultiRoom`/#182 on the same `room_grid` primitives. Didn't build the shared BFS-reachability structural-check primitive you suggested in this PR, since the fix itself (excluding blocker neighbourhoods) resolves the concrete case here; happy to split that out as a follow-up issue if useful for future room-grid work.

PR #201, opened 2026-09-03
feat(environments): implement MultiRoom (#182)

### LINKED ISSUE #182: Implement MultiRoom family (N2-S4, N4-S5, N6)

epignatelli:
Closes #182.

## Summary

Ports MiniGrid's `MultiRoomEnv`: `n` rooms chained by unlocked doors,
start in room 0, goal in room `n-1`. Registers 3 variants matching
MiniGrid's own kwargs:

- `Navix-MultiRoom-N2-S4-v0` (2 rooms, max size 4)
- `Navix-MultiRoom-N4-S5-v0` (4 rooms, max size 5 - MiniGrid's `-v1`
  kwargs; its own `-v0` is a documented legacy bug actually configured
  for 6 rooms, kept upstream for backwards compatibility - not ported
  here, per this repo's "port `v1`, name it `v0`, document the
  difference" convention)
- `Navix-MultiRoom-N6-v0` (6 rooms, max size 10)

## Why this took real design work

MiniGrid's own generation is retry/backtrack-based (`_placeRoom`, up
to 8 retries per room, unbounded chain restarts on failure) - not
directly JAX-traceable. This is kept faithful to the real algorithm
rather than approximated: bounded retry via nested
`jax.lax.while_loop` (per-room placement retry, then whole-chain
restart on failure), with a guaranteed-valid deterministic fallback
layout selected via `jnp.where` for the rare case the random search
doesn't converge within its retry budget - so the environment is
provably 100% correct by construction, not just "usually correct".

All 3 registrations use MiniGrid's own fixed 25x25 grid - its retry
logic is what keeps a random layout compact, not an adaptively-sized
canvas.

`N6` is the heaviest environment in navix by a wide margin (documented
in `multi_room.py`'s module docstring) and is deliberately excluded
from the test suite's gameplay walkthrough - structural + jit/vmap
checks only - to keep CI time bounded.

## Bugs found and fixed during verification

Both found via direct seed-by-seed tracing against `_generate_layout`'s
raw output, not guesswork:

1. **Room-to-room connectivity**: the wall-relative room-positioning
   formulas sampled the perpendicular offset from a half-open range
   that, due to `jax.random.randint`'s exclusive `high`, accidentally
   *included* the door's own coordinate - letting a door land on the
   new room's corner instead of its interior, disconnecting the two
   rooms while still passing every overlap/bounds check (so it wasn't
   caught by those). Fixed by excluding that endpoint, matching
   MiniGrid's own `[coord-size+2, coord)` range exactly.
2. **Fallback layout off-grid**: the deterministic fallback chain
   centred only its first room instead of the chain's total span,
   running the last room off the grid edge for larger room counts.

Also fixed a latent performance trap in the test's own BFS-navigation
helpers: calling `env.step` un-jitted in a loop retraces and
recompiles the *entire* autoreset-embedded reset (`Environment.step`'s
`jax.lax.cond` branch) on every call, since eager `lax.cond` doesn't
cache across separate Python calls - harmless for envs with a cheap
reset, but for `MultiRoom`'s genuinely heavy nested-retry reset this
measurably balloons to minutes of compile time and multiple GB before
the process runs out of memory. Fixed by jitting the step function
once per env and reusing it across the whole BFS walk.

## Verification

- Real `env.reset`/`env.step` execution (not just compilation) across
  seeds for all 3 registrations.
- Full BFS-navigated gameplay walkthrough (open every door, reach the
  goal, confirm termination + positive reward) for N2/N4.
- `tests/test_multi_room.py`: `3 passed` (structure, gameplay +
  reward/termination, jit/vmap) in ~2m13s.
- `mypy`/`pylint` on the changed files clean modulo pre-existing/
  unavoidable patterns already accepted elsewhere in the codebase.

### DIFF
--- a/navix/environments/__init__.py
+++ b/navix/environments/__init__.py
@@ -38,3 +38,4 @@ from .fetch import Fetch
 from .put_near import PutNear
 from .memory import Memory
 from .obstructed_maze import ObstructedMaze1Dlhb, ObstructedMazeFull
+from .multi_room import MultiRoom
--- a/navix/environments/multi_room.py
+++ b/navix/environments/multi_room.py
@@ -46,7 +46,21 @@ set - so `N6` is genuinely, permanently more expensive to run than any
 other navix environment, not just slower to compile. Kept anyway,
 deliberately, for fidelity - see this session's own discussion on
 issue #182 for the (rejected) cheaper alternative (a straight room
-chain) and the actual cost numbers."""
+chain) and the actual cost numbers.
+
+One consequence worth calling out explicitly: `Environment.step`
+embeds a full `Environment.reset` call as its `jax.lax.cond` autoreset
+branch, so calling `env.step(...)` *un-jitted* in a Python loop
+retraces and recompiles that whole branch - reset included - on every
+single call, since eager `lax.cond` doesn't cache across separate
+top-level calls the way `jax.jit` does. For most navix environments
+that's an unnoticeable cost, because their `reset` is cheap. For
+`MultiRoom`, and especially `N6`, `reset` is exactly the heavy nested-
+retry search described above, so this pattern is a real footgun: it
+was measured to balloon to minutes of compile time and multiple GB of
+memory before crashing outright. Always `jax.jit` (or `jax.vmap` over
+a jitted function) `env.step`/`env.reset` before calling either in a
+loop against any `MultiRoom` variant."""
 
 from __future__ import annotations
 from typing import List, Tuple, Union
@@ -69,7 +83,7 @@ from .registry import register_env
 
 GRID_SIZE = 25  # fixed for every registration, matching MiniGrid's own
 MIN_ROOM_SIZE = 4  # matches MiniGrid's hardcoded minSz
-BOUNDARY_MARGIN = MIN_ROOM_SIZE - 1  # see _place_room's docstring
+BOUNDARY_MARGIN = MIN_ROOM_SIZE - 1  # see place_room's docstring
 MAX_PLACEMENT_TRIES = 64  # MiniGrid itself retries only 8 times per room,
 # but a JAX retry (one more jax.lax.while_loop iteration, already-compiled)
 # has essentially none of the real cost a Python-level retry does - bumped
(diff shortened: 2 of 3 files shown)


epignatelli:
@claude please review

epignatelli:
Addressed in 678e141:

1. **Missing fallback-layout test coverage** — added `test_multi_room_fallback_layout_valid` (direct unit test of `fallback_layout` for every registration's room count) and `test_multi_room_fallback_path_forced` (monkeypatches `MAX_PLACEMENT_TRIES` to 0 so `_reset`'s `jnp.where(valid, ..., fallback...)` swap is exercised through a real `env.reset` call, per your suggested approach). Both also caught by tightening the shared structural check to verify every door actually borders floor on both sides — a pure bounds/overlap check can't catch a door landing on a room's own corner.
2. **N6 step-jit footgun** — now documented in `multi_room.py`'s own module docstring, not just test-file comments.
3. **`max_steps` dead-code pattern** — agreed this is a pre-existing, systemic `registry.make()` issue shared with `memory.py`/`obstructed_maze.py`, not specific to this PR; left as a follow-up per your own note.

Also renamed every module-level helper this PR introduced to drop leading underscores (`_place_room` → `place_room`, etc.) — an in-session convention this file missed, unrelated to your review but caught while touching the file.

`pytest tests/test_multi_room.py`: 5 passed in 161s.

Issue #10, opened 2023-06-14
Implement more environments


epignatelli:
Source:  https://minigrid.farama.org/environments/minigrid/

- [x] MiniGrid-BlockedUnlockPickup-v0
- [x] MiniGrid-LavaCrossingS9N1-v0
- [x] MiniGrid-LavaCrossingS9N2-v0
- [x] MiniGrid-LavaCrossingS9N3-v0
- [x] MiniGrid-LavaCrossingS11N5-v0
- [x] MiniGrid-SimpleCrossingS9N1-v0
- [x] MiniGrid-SimpleCrossingS9N2-v0
- [x] MiniGrid-SimpleCrossingS9N3-v0
- [x] MiniGrid-SimpleCrossingS11N5-v0
- [x] MiniGrid-DistShift1-v0
- [x] MiniGrid-DistShift2-v0
- [x] MiniGrid-DoorKey-5x5-v0
- [x] MiniGrid-DoorKey-6x6-v0
- [x] MiniGrid-DoorKey-8x8-v0
- [x] MiniGrid-DoorKey-16x16-v0
- [x] MiniGrid-Dynamic-Obstacles-5x5-v0
- [x] MiniGrid-Dynamic-Obstacles-Random-5x5-v0
- [x] MiniGrid-Dynamic-Obstacles-6x6-v0
- [x] MiniGrid-Dynamic-Obstacles-Random-6x6-v0
- [x] MiniGrid-Dynamic-Obstacles-8x8-v0
- [x] MiniGrid-Dynamic-Obstacles-16x16-v0
- [x] MiniGrid-Empty-5x5-v0
- [x] MiniGrid-Empty-Random-5x5-v0
- [x] MiniGrid-Empty-6x6-v0
- [x] MiniGrid-Empty-Random-6x6-v0
- [x] MiniGrid-Empty-8x8-v0
- [x] MiniGrid-Empty-16x16-v0
- [x] MiniGrid-Fetch-5x5-N2-v0
- [x] MiniGrid-Fetch-6x6-N2-v0
- [x] MiniGrid-Fetch-8x8-N3-v0
- [x] MiniGrid-FourRooms-v0
- [x] MiniGrid-GoToDoor-5x5-v0
- [x] MiniGrid-GoToDoor-6x6-v0
- [x] MiniGrid-GoToDoor-8x8-v0
- [x] MiniGrid-GoToObject-6x6-N2-v0
- [x] MiniGrid-GoToObject-8x8-N2-v0
[...]

PR #204, opened 2026-09-03
feat(environments): implement LockedRoom and Playground (#179, #184)

### LINKED ISSUE #179: Implement LockedRoom
Part of #10.

## Environment
- `MiniGrid-LockedRoom-v0`

## Description
A larger single-floor layout: a central room plus four side rooms arranged in a cross, one side room locked (its key hidden randomly in one of the *other* rooms). Reward + termination on reaching a `Goal` inside the locked room. MiniGrid's own `LockedRoom` doesn't use `RoomGrid` (unlike `Unlock`/`UnlockPickup`/`ObstructedMaze`) - it hand-rolls the cross layout directly.

## Effort: medium
Structurally closer to `navix`'s existing `KeyCorridor` (`navix/environments/key_corridor.py`, already implements "locked room, key hidden elsewhere, reach a goal past the lock") than to the `RoomGrid`-based families - likely reuses `KeyCorridor`'s generation approach and patterns more directly than #174 (RoomGrid) does, though the cross-shaped (not linear-corridor-shaped) room arrangement is genuinely different and needs its own layout logic. Worth comparing both approaches carefully before choosing one to extend.

---

### LINKED ISSUE #184: Implement Playground
Part of #10.

## Environment
- `MiniGrid-Playground-v0`

## Description
An open room scattered with random walls, doors and objects, with no goal, no reward function, and no termination condition beyond timeout - used in the original MiniGrid/Minigrid paper for visualisation and free-form interaction, not as a solvable benchmark.

## Effort: trivial, but flagging low priority
Implementation itself would be low effort (random object/wall scattering, reusing entities that all already exist), but it has no differentiated value as an RL benchmark - nothing to solve, no reward signal, so it wouldn't be used in any of navix's actual benchmark suites (`navix/benchmarks/`) the way every other environment in #10 would be. Suggest deprioritising this below every other family unless someone specifically wants it for rendering/visualisation demos.

---

epignatelli:
_Posted by an AI agent (Claude Code) on behalf of @epignatelli._

Closes #179, closes #184 - both part of #10's environment checklist, implemented together per request.

## `LockedRoom` (#179, `MiniGrid-LockedRoom-v0`)

Six rooms - three stacked on each side of a central north-south hallway - each opening onto that hallway through its own door. One room is locked; its `Goal` can only be reached by finding a `Key` of the matching colour hidden in one of the other five rooms, then unlocking that one door.

The six rooms' positions/sizes/door cells are a fixed function of `height`/`width` (verified directly against MiniGrid's actual `LockedRoomEnv._gen_grid` source - `left_wall`/`right_wall`/`room_w`/`room_h`/the `j + 3` per-room door-row offset all reproduced exactly). Only which room is locked, the six door colours, which room hides the key, and every entity's position within its own room are randomised per episode - no `jax.lax.while_loop` retry search needed anywhere, unlike `MultiRoom` (#182).

## `Playground` (#184, `MiniGrid-Playground-v0`)

A 3x3 grid of rooms, every orthogonally-adjacent pair joined by one unlocked, randomly-positioned, randomly-coloured door. 12 objects, each independently `Key`/`Ball`/`Box` (matching MiniGrid's own per-object random type draw), are scattered across the grid - reusing `GoToObject`'s `DISCARD_PILE_COORDS` fixed-slot-allocation trick for the variable per-type counts.

No goal, no reward beyond a flat 0, no termination beyond the timeout. Registers with `rewards.free` (see below) and `termination_fn=terminations.compose()` (i.e. never terminates) rather than the default `DEFAULT_TERMINATION` - the latter's `on_ball_hit` turned out to fire from an ordinary player-bumps-`Ball` collision (`actions._can_walk_there`'s generic `record_walk_into` dispatch), not just a moving ball colliding with the player, and ended episodes within a handful of random steps in testing before this was caught. See the module docstring for the full trace of that finding.

## Drive-by fix: `rewards.free`'s signature

`rewards.free` was defined as `free(state) -> Array`, but every `reward_fn` is called as `self.reward_fn(prev_state, action, state)` (`Environment._step`) - a 3-argument call. It had never actually been wired to any registration, so this was unreachable dead code until Playground's `reward_fn=rewards.free` needed it. Fixed as a separate, first commit; zero behaviour change to any existing caller since none exist.

## Testing

New `tests/test_locked_room.py` / `tests/test_playground.py`: structural checks against the live reset state (room/door layout, distinct colours, locked-room/key-room invariants for LockedRoom; door/object counts and mutual distinctness for Playground), a full real-gameplay walkthrough for LockedRoom (open the key's own door, fetch the key, unlock the locked door, reach the goal), a reward-free/no-early-termination random walk for Playground, and jit/vmap compatibility for both.

```
$ uv run pytest tests/test_locked_room.py tests/test_playground.py tests/test_registry.py tests/test_environments.py -q
16 passed
```


epignatelli:
@claude please revierw

epignatelli:
Pushed a follow-up commit addressing a question in review: `NotImplementedEnvs` turned out to be far more stale than just the two entries this PR's own environments removed.

30 of its 34 entries already had a matching `Navix-*` registration from past PRs that never pruned their own MiniGrid id here (`BlockedUnlockPickup`, `LavaCrossing*`, `Fetch*`, `GoToObject*`, `Memory*`, `MultiRoom*`, the `ObstructedMaze` `-v0` ids, `PutNear*`, `RedBlueDoors*`, `Unlock*`) - pruned all of them.

Kept the 4 `ObstructedMaze` `-v1` ids (`2Dlhb`/`1Q`/`2Q`/`Full`) deliberately: navix only ports MiniGrid's `-v1` fix for these, registered under the `-v0` id, so the literal `-v1` MiniGrid id has no registration of its own - still fair to call it unimplemented. To make that aliasing visible where it actually matters, those 4 `-v0` ids now emit a `UserWarning` on `nx.make(...)` naming both MiniGrid ids and the behavioural difference (`2Dl`/`2Dlh`, which port MiniGrid's `-v0` directly, don't warn).

epignatelli:
Correcting my previous comment - I'd implemented the wrong trigger the first time. Force-pushed a fix.

**What actually happens now:** `nx.make("MiniGrid-ObstructedMaze-2Dlhb-v1")` (the literal MiniGrid id) redirects to `Navix-ObstructedMaze-2Dlhb-v0` and warns that it's doing so, instead of raising `NotImplementedError`. Same for `1Q`/`2Q`/`Full`, plus `MiniGrid-MultiRoom-N4-S5-v1` (same v1-becomes-v0 situation, not previously in `NotImplementedEnvs` at all). Using the actual `Navix-*-v0` id directly still doesn't warn - only the MiniGrid `-v1` alias does.

```python
>>> nx.make("MiniGrid-ObstructedMaze-2Dlhb-v1")
UserWarning: MiniGrid-ObstructedMaze-2Dlhb-v1 has no distinct navix registration -
Navix-ObstructedMaze-2Dlhb-v0 already implements MiniGrid's own `-v1` behaviour for
this environment family (see V1_ALIASES in registry.py). Instantiating
Navix-ObstructedMaze-2Dlhb-v0 instead.
```

`NotImplementedEnvs` is now empty (all 30 stale entries pruned, the remaining 4 handled via the alias above instead of just left there) - kept as a named list for the next real gap.

New `test_v1_aliases_redirect_and_warn` covers all 5 aliases: warns exactly once, names the target id, constructs the same environment as calling the target directly, and the target itself stays silent. `test_registry.py`/`test_environments.py`/`test_obstructed_maze.py`/`test_multi_room.py`: 20 passed.

PR #205, opened 2026-09-03
refactor(events): standardize events.py to (prev_state, action, state) (#192)

### LINKED ISSUE #192: Standardize events.py function signatures to (prev_state, action, state)
Surfaced during PR #191's work: `events.py` currently has 5 different function signature shapes across its 18 functions:

- `(state)` - 12 functions (the original convention: `on_goal_reached`, `on_lava_fall`, `on_ball_hit`, `on_door_done`, `on_door_open`, `on_box_pickup`, `on_wall_hit`, `on_key_pickup`, `on_ball_pickup`, `on_target_fetched`, `on_any_target_pickup`, `on_put_near_wrong_pickup`)
- `(prev_state, state)` - 2 functions (`on_ordered_doors_success`/`on_ordered_doors_failure`)
- `(action, state)` - 1 function (`on_target_done`)
- `(action)` - 1 function (`on_wrong_toggle`)
- `(prev_state, action)` - 1 function (`on_put_near_drop_attempted`)
- `(prev_state, action, state)` - 1 function (`on_put_near_success`)

Each of the non-`(state)` ones needed its own docstring paragraph explaining why it deviates from the "established" single-state convention.

## The idea

An event is conceptually always a check over one transition - `(prev_state, action, state)` - even when a specific check only reads a subset of that triple internally. `rewards.py`/`terminations.py`'s own wrapper functions *already* uniformly take `(prev_state, action, state)`, even when their body only touches `state` (e.g. `rewards.on_goal_reached(prev_state, action, state)` ignores the first two args entirely) - `events.py` just never adopted the same convention.

Standardizing every `events.py` function to `(prev_state, action, state)` unconditionally (ignoring unused args in the body, same as the wrapper layer already does) would:
- Give one honest, uniform contract instead of five ad-hoc ones.
- Let `rewards.py`/`terminations.py` wrappers become trivial, uniform pass-throughs (right now some forward 1 arg, some 2, some 3).
- Remove every "this deviates from the convention, here's why" docstring caveat added this PR.

## Scope (surveyed, not yet done)

- 18 function signatures in `events.py`.
- 2 internal `events.py` call sites (`on_any_target_pickup`'s OR-chain, `on_put_near_wrong_pickup`'s call into it).
- ~21 `rewards.py`/`terminations.py` wrapper bodies.
- 9 direct external call sites (`tests/test_events.py`, `tests/test_unlock.py`) calling `nx.events.on_X(state)` directly with the old single-arg form - each needs a `prev_state`/`action` added; checked one (`test_grid_hit_boundary_not_wall_entity`) and it has a natural prior state available a line earlier, so this should be mechanical rather than needing dummy placeholders, but worth confirming for the rest when this is picked up.

Deferred out of PR #191 to keep that PR scoped to the 5 environments it's actually implementing - purely a refactor, no behavior change intended.

epignatelli:
Closes #192.

## events.py signature standardization

`events.py` had 5 different function signature shapes spread across its
20 functions - `(state)`, `(prev_state, state)`, `(action, state)`,
`(action)`, `(prev_state, action)` - each deviation needing its own
docstring paragraph explaining why it broke from the "established"
single-state convention.

`rewards.py`/`terminations.py`'s own wrapper functions already take a
uniform `(prev_state, action, state)`, even when a given check only
reads `state` (e.g. `rewards.on_goal_reached` ignores the first two
args entirely) - `events.py` just never adopted the same convention.

This standardizes every function in `events.py` to that same triple.
Pure refactor, no behavior change - unused args are still named in
each function's docstring, marked `(unused)`.

- `rewards.py`/`terminations.py`: every `events.on_*` call site now
  forwards the full triple instead of 1, 2, or 3 args ad hoc.
- `tests/test_events.py`/`tests/test_unlock.py`: the 8 direct
  `nx.events.on_*` call sites updated to pass the full triple - a
  natural `prev_state` was available a line or two earlier at every
  site except `test_events.py`'s bare `navix.actions.*` calls (not
  `env.step`, so no integer action id is ever in scope there), which
  use a documented `UNUSED_ACTION` placeholder instead.

## .test_durations regeneration

Also included: a `.test_durations` regeneration (162 -> 186 tests),
since several environments (`Memory`, `MultiRoom`, `ObstructedMaze`'s
1D/1Q/2Q/Full variants) landed since the last regeneration without
duration entries. Same `pytest --store-durations -n 2` settings as the
last regeneration, run on bologna under a Python 3.14.7 env matching
CI's pin. 9/186 tests hit the same LLVM-JIT/`vm.max_map_count`
resource-pressure crash already documented for this shared machine in
an earlier commit - not a real regression (none touch this PR's own
`events.py`/`rewards.py`/`terminations.py` changes, and
`tests/test_events.py` itself passed clean); see the commit message
for the full verification.


epignatelli:
@claude please review

PR #207, opened 2026-09-03
feat(agents): transformer frame-history encoder for POMDP tasks (#169), Option 1 — stored carry

### LINKED ISSUE #169: Transformer-based encoder for POMDP tasks encoding a history of frames
## Problem

Every navix agent's `pomdp` observation mode (`rgb_first_person`/`categorical_first_person`/`symbolic_first_person` in `navix/observations.py`) returns a single current-step, first-person frame - no history. That's not actually a POMDP-compliant observation: a lone frame doesn't disambiguate states that look identical from the agent's current viewpoint but differ in what led there (e.g. which direction something moved before this frame), so the policy only sees a Markovian *approximation* of the true partially-observable state, not a proper belief/history-conditioned one.

## Proposal

Add a transformer-based encoder (alongside the existing `ConvEncoder`/`QConvEncoder`/`MLPEncoder`/`QMLPEncoder` family in `navix/agents/models.py`) that attends over the last `c` observed frames rather than just the current one, so `pomdp`-mode agents get an observation that's actually consistent with the POMDP definition (a short window of history standing in for the missing state, rather than a single frame pretending to be Markovian).

The user pointed to https://arxiv.org/pdf/2608.07335 as a possibly-useful reference here - worth flagging that its actual abstract/keywords ("Aftab: A Comprehensive Benchmark of CNN Encoders and Advanced Value Functions in Parallelized Q-Networks") describe a CNN-encoder/value-head benchmark for PQN specifically, not a transformer/frame-history encoder, so it may not be the intended paper or may only be tangentially related - worth double-checking the reference before using it to guide the design.

## Scope (not yet decided, needs its own design pass)

- Where the `c`-frame history buffer lives (agent training state vs. environment/observation-function level) and how it interacts with navix's existing autoreset/episode-boundary handling.
- Whether this is a new observation function (`rgb_first_person_history`-style) or a new encoder that itself manages a frame window from a single-frame observation function.
- Compute/memory cost of attention over `c` frames at navix's usual `num_envs`/`num_steps` scale.

---

epignatelli:

Closes #169.

## What this does

Generalises every PPO/PQN feature extractor to a **carry-based `Encoder` contract** so the MDP↔POMDP distinction lives entirely in the encoder — you swap the encoder at `ActorCritic(...)` / `QNetwork(...)` construction and nothing else in the agent, the environment, or the observation function changes:

```python
class Encoder(nn.Module):
    def initial_carry(self, obs_shape) -> carry: ...     # () default (stateless)
    def __call__(self, carry, obs, is_first) -> (carry, features): ...
```

- `MLPEncoder` / `ConvEncoder` / `QMLPEncoder` / `QConvEncoder` are stateless `Encoder` subclasses: carry `()`, ignore `is_first`, features unchanged. **Fixed-seed MDP runs of both PPO and PQN are byte-identical to `main`.**
- `TransformerEncoder` (+ `TransformerBlock`) is the stateful one (#169): its carry is the raw `(context, *frame)` window of the last `context` observations. The agent threads the carry like `env_state`; the encoder resets its window on `is_first` (`t == 0`, robust to navix's one-step-deferred autoreset).
- `ActorCritic` / `QNetwork` thread **one shared carry** to their encoder(s) and return `(carry, ...)` from every method.
- Dreamer's `Encoder` is renamed `SymlogEncoder` (symlog-input RSSM encoder — different `__call__` shape, no carry; internal, not re-exported).

Both agents use **Option 1** (stored carry): the per-step window is cached in `Buffer` and replayed in the loss. Because the window is raw frames (no dependence on encoder parameters) this is **exact**, not an approximation — verified against the Option 2 alternative (#208), which produces the same gradient at ~4x the loss cost and is only worth it for a future *learned* recurrent encoder.

## Commits

1. `feat` — carry contract + `TransformerEncoder`, wired through PPO
2. `test` — encoder-contract + PPO-wiring coverage; `examples/ppo_pomdp.py`
3. `fix` — `ActorCritic` threads one shared carry (a per-head tuple left the critic's window frozen, since `collect_experience` only runs `policy`)
4. `style` — uniform `__call__(carry, obs, is_first)` signature
5. `refactor` — rename Dreamer's `Encoder` → `SymlogEncoder`
6. `refactor` — `Encoder` base class for the contract
7. `feat` — thread the carry through PQN

## Not in scope

- Dreamer (its RSSM is already history-conditioned).
- A conv/ViT patch-embed `frame_encoder` for `rgb_first_person` (flatten+Dense per frame is the first cut).
- `FrameStack`-style env wrappers / new observation functions — the encoder owns history instead.

## Tests / examples

- `tests/test_models.py` — the `Encoder` base, contract, `TransformerEncoder` window reset / rolling / history sensitivity / shared per-frame weights / jit+vmap.
- `tests/test_ppo.py`, `tests/test_pqn.py` — each agent trains one update without NaNs with a stateless encoder (carry `()` end to end) and with `TransformerEncoder` (attention params present, `Buffer.carry` is the window).
- `examples/ppo_pomdp.py` — PPO on a first-person observation, identical to `examples/ppo.py` except the encoder.

## Behavior change

None for existing environments or existing PPO/PQN configs — the stateless-encoder path is byte-identical to `main` for both agents.

### DIFF
--- a/examples/ppo_pomdp.py
+++ b/examples/ppo_pomdp.py
@@ -1,0 +1,76 @@
+"""Same `PPO` as `examples/ppo.py`, on a first-person (`pomdp`)
+observation - the only change is the encoder. A single first-person frame
+is not a Markovian observation (issue #169); `TransformerEncoder` gives
+the policy a short window of history instead, and it manages that window
+itself (as its carry), so the agent, the environment, and the observation
+function are all identical to the fully-observable case.
+"""
+
+from dataclasses import dataclass, field
+import tyro
+import numpy as np
+import jax.numpy as jnp
+import navix as nx
+from navix import observations
+from navix.agents import PPO, PPOHparams, ActorCritic, MLPEncoder, TransformerEncoder
+from navix.environments.environment import Environment
+
+
+@dataclass
+class Args:
+    project_name = "navix-examples"
+    seeds_offset: int = 0
+    n_seeds: int = 1
+    env_id: str = "Navix-DoorKey-Random-6x6-v0"
+    discount: float = 0.99
+    context: int = 4
+    """Number of frames the encoder attends over."""
+    ppo_config: PPOHparams = field(default_factory=lambda: PPOHparams(budget=5_000))
+    """Small budget - a quick usage demo, not a convergence benchmark."""
+
+
+if __name__ == "__main__":
+    args = tyro.cli(Args)
+
+    def FlattenObsWrapper(env: Environment):
+        # per-frame flatten: the encoder attends over a window of these.
+        flatten_obs_fn = lambda x: jnp.ravel(env.observation_fn(x))
+        flatten_obs_shape = (int(np.prod(env.observation_space.shape)),)
+        return env.replace(
+            observation_fn=flatten_obs_fn,
+            observation_space=env.observation_space.replace(shape=flatten_obs_shape),
+        )
+
+    env = nx.make(
+        args.env_id,
+        observation_fn=observations.symbolic_first_person,
+        gamma=args.discount,
+    )
+    env = FlattenObsWrapper(env)
+
+    def encoder() -> TransformerEncoder:
+        return TransformerEncoder(
+            frame_encoder=MLPEncoder(hidden_size=64),
+            hidden_size=64,
+            context=args.context,
+        )
+
+    agent = PPO(
+        hparams=args.ppo_config,
+        # the ONLY difference from examples/ppo.py: the encoder.
+        network=ActorCritic(
+            action_dim=len(env.action_set),
+            actor_encoder=encoder(),
+            critic_encoder=encoder(),
+        ),
+        env=env,
+    )
+
+    experiment = nx.Experiment(
+        name=args.project_name,
+        agent=agent,
+        env=env,
+        env_id=args.env_id,
+        seeds=tuple(range(args.seeds_offset, args.seeds_offset + args.n_seeds)),
+    )
+    train_state, logs = experiment.run()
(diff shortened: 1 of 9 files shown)


epignatelli:
@claude please do a very very thorough review

epignatelli:

Addressed in `a80e367`:

- **float32 carry / memory (New risk 1).** `initial_carry` now takes an optional `dtype` (default `float32`, ignored by the stateless `()` default); `PPO`/`PQN` pass `init_x.dtype`, so for navix's `uint8` first-person observations the `(context, *frame)` window is stored `uint8` throughout `Buffer.carry` (4x smaller — confirmed `Buffer.carry.dtype == uint8` in a real rollout). The per-frame `frame_encoder` is fed `float32` (cast at the call site, not in the stored window).
- **Mismatched actor/critic carries (Requested changes).** `ActorCritic.initial_carry` now raises `ValueError` with actor/critic carry shapes if they disagree, instead of failing as an opaque trace-time shape error.
- **Learned-carry gap (Broader opportunities).** `Encoder`'s docstring now states that the store-and-replay carry handling is exact only for a parameter-independent carry, and a learned recurrent carry (GRU/SSM) needs per-sequence recomputation in the loss.
- **BREAKING CHANGE footer (Breaking changes).** Deliberately not adding one — `ActorCritic`/`QNetwork` are internal building blocks assembled by `PPO`/`PQN`, not a documented end-user entry point.

New tests cover the uint8 window and the mismatch `ValueError`; MDP runs of both PPO and PQN remain byte-identical to `main`.

PR #208, opened 2026-09-03
feat(agents): transformer frame-history encoder for POMDP tasks (#169), Option 2 — sequence rescan

### LINKED ISSUE #169: Transformer-based encoder for POMDP tasks encoding a history of frames

epignatelli:

Closes #169 (one of two alternative implementations — see #207 for Option 1).

## What this does

Same **carry-based encoder contract** and `TransformerEncoder` as #207 (shared
commits) — see that PR for the framework description. The difference is entirely
in how `PPO`'s loss obtains the encoder carry.

## Option 2: sequence rescan

Instead of caching a per-step carry in `Buffer`, `PPO` keeps only the
rollout-initial carry (`TrainingState.rollout_carry`) and **rescans the encoder
over each whole rollout sequence** in the loss (`forward_sequence`, a
`jax.lax.scan` over `num_steps`). Gradients flow through the recurrence (full
BPTT), so this generalises to a future *learned*-recurrent encoder (GRU/SSM),
not just `TransformerEncoder`'s param-independent frame window.

For `TransformerEncoder` specifically this produces the **same update** as #207
(verified: a single full-batch gradient step agrees to ~7 significant figures —
the residual is float non-associativity between a vmapped `T·N` reduction and a
`scan`-over-`T` reduction).

| | Option 1 (#207) | Option 2 (this PR) |
|---|---|---|
| `ppo.py` diff | +98 / −52 | +208 / −80 |
| minibatching | unchanged (shuffled `T·N`) | over env axis; `num_minibatches` must divide `num_envs` |
| MDP path vs `main` | byte-identical | **changed** (see below) |
| per-update wall time (CPU, ctx=4) | ~142 ms | ~538 ms (3.8×) |
| rollout-buffer memory | `c×` (stores the window) | 1× (single frames) + BPTT activations |
| supports a future learned-recurrent encoder | no | yes (full BPTT) |

## Behavior change (flagged per AGENTS.md)

- **`num_minibatches` semantics.** Minibatches are now groups of whole rollout
  sequences over the **env** axis, so `num_minibatches` must divide `num_envs`
  (was `num_steps * num_envs`). PPO's own defaults (16 envs, 8 minibatches)
  still divide; a config with `num_minibatches > num_envs` (e.g. PQN-style 32)
  needs changing.
- **MDP (stateless-encoder) PPO is no longer numerically identical to `main`**
  for a fixed seed — the minibatch structure differs (fewer, larger,
  temporally-contiguous minibatches). Per-sample loss math is unchanged;
  optimisation dynamics shift. No `BREAKING CHANGE:` footer because no public
  signature/format changes — but a fixed-seed benchmark curve will move.

## Tests / examples

Same as #207, adapted: `tests/test_ppo.py` checks `forward_sequence` rescans a
real rolling window and that `rollout_carry` tracks the per-env window;
`examples/ppo_pomdp.py` runs clean.

### DIFF
--- a/examples/ppo_pomdp.py
+++ b/examples/ppo_pomdp.py
@@ -1,0 +1,76 @@
+"""Same `PPO` as `examples/ppo.py`, on a first-person (`pomdp`)
+observation - the only change is the encoder. A single first-person frame
+is not a Markovian observation (issue #169); `TransformerEncoder` gives
+the policy a short window of history instead, and it manages that window
+itself (as its carry), so the agent, the environment, and the observation
+function are all identical to the fully-observable case.
+"""
+
+from dataclasses import dataclass, field
+import tyro
+import numpy as np
+import jax.numpy as jnp
+import navix as nx
+from navix import observations
+from navix.agents import PPO, PPOHparams, ActorCritic, MLPEncoder, TransformerEncoder
+from navix.environments.environment import Environment
+
+
+@dataclass
+class Args:
+    project_name = "navix-examples"
+    seeds_offset: int = 0
+    n_seeds: int = 1
+    env_id: str = "Navix-DoorKey-Random-6x6-v0"
+    discount: float = 0.99
+    context: int = 4
+    """Number of frames the encoder attends over."""
+    ppo_config: PPOHparams = field(default_factory=lambda: PPOHparams(budget=5_000))
+    """Small budget - a quick usage demo, not a convergence benchmark."""
+
+
+if __name__ == "__main__":
+    args = tyro.cli(Args)
+
+    def FlattenObsWrapper(env: Environment):
+        # per-frame flatten: the encoder attends over a window of these.
+        flatten_obs_fn = lambda x: jnp.ravel(env.observation_fn(x))
+        flatten_obs_shape = (int(np.prod(env.observation_space.shape)),)
+        return env.replace(
+            observation_fn=flatten_obs_fn,
+            observation_space=env.observation_space.replace(shape=flatten_obs_shape),
+        )
+
+    env = nx.make(
+        args.env_id,
+        observation_fn=observations.symbolic_first_person,
+        gamma=args.discount,
+    )
+    env = FlattenObsWrapper(env)
+
+    def encoder() -> TransformerEncoder:
+        return TransformerEncoder(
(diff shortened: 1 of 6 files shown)


epignatelli:
Closing in favour of #207 (Option 1 — stored carry), which was chosen: for a frame-window (parameter-independent) carry the two produce the same gradient, and Option 1 is the smaller change (MDP path byte-identical, minibatching untouched). This sequence-rescan branch stays as a reference for if a future *learned* recurrent encoder (GRU/SSM) needs full BPTT.

PR #218, opened 2026-09-04
fix(rewards): deprecate action_cost, drop per-step cost from DEFAULT_TASK (#211)

### LINKED ISSUE #211: `rewards.action_cost` hard-codes the action index and mis-describes it
_Filed by an AI agent on behalf of @epignatelli, from findings during the #167 documentation pass (see PR #209)._

```python
def action_cost(prev_state, action, new_state, cost=0.01):
    # noops are free
    return -jnp.asarray(action != 6, dtype=jnp.float32) * cost
```

Two problems:

1. **Wrong premise.** Index `6` in the default `MINIGRID_ACTION_SET` is `done`, not `noop` (`noop` isn't in that set at all). So the "free" action is actually `done`, and the "noops are free" comment/docstring is misleading.
2. **Fragile.** The literal `6` only lines up with `MINIGRID_ACTION_SET`. Under `COMPLETE_ACTION_SET`, index `6` is `left` (so strafing is free) and `done` is index `9` (so it's penalised). `navix.events` already solved this for itself with `DONE_ACTION = MINIGRID_ACTION_SET.index(actions.done)` — `action_cost` should do the same, or take the exempt action(s) as a parameter.

`DEFAULT_TASK` composes `action_cost`, so this affects every default environment.

epignatelli:

Closes #211.

## `action_cost` → deprecated alias of `time_cost`

`action_cost` exempted the action at index `6` (`done` in `MINIGRID_ACTION_SET`) from the per-step cost. But a reward function's signature is `(prev_state, action, state)` — it never sees the environment's `action_set`, so the hard-coded `6` is wrong for any other set (it's `left` in `COMPLETE_ACTION_SET`, out of range for a 3-action `DynamicObstacles`). There's no way to make an index-based exemption correct here.

With the exemption gone it's identical to `time_cost`, so `action_cost` now **emits a `DeprecationWarning` and delegates to `time_cost`**. Same name/signature.

## `DEFAULT_TASK` — no per-step shaping

`DEFAULT_TASK` was `compose(on_goal_reached, action_cost)`; it's now just **`on_goal_reached`** (`+1` at goal, `0` otherwise). `gamma` (0.99) is unchanged.

**Scope check** — instantiated all 83 registered envs: `multi_room.py` was the **only** one that opted into `DEFAULT_TASK` (`reward_fn=kwargs.pop("reward_fn", rewards.DEFAULT_TASK)`). To keep `Navix-MultiRoom-{N2-S4,N4-S5,N6}-v0` unchanged it now pins `compose(on_goal_reached, time_cost)` explicitly. Every other env already passes its own `reward_fn`. So **no `Navix-*` id's reward changes**; the `BREAKING CHANGE:` footer covers only the `Environment.create` fallback (`0.0` instead of `-0.01` on non-goal steps) and the `action_cost` deprecation.

## Also

- `empty.py`'s `Room` docstring said "…the goal minus a per-step cost" — already wrong (`Navix-Empty-*` uses `on_goal_reached`), fixed.
- Tests: `action_cost` warns + equals `time_cost`; `DEFAULT_TASK is on_goal_reached` and gives `0.0` on a non-goal step; `Navix-MultiRoom-N2-S4-v0` still gives `-0.01` on a non-goal step; `test_tasks_composition` off the deprecated fn.

Verified: `tests/test_tasks.py`, `tests/test_multi_room.py`, `tests/test_environments.py`, `tests/test_events.py` pass; `examples/ppo.py` + `examples/hparam_search.py` run clean.

### DIFF
--- a/navix/environments/empty.py
+++ b/navix/environments/empty.py
@@ -45,8 +45,8 @@ from .registry import register_env
 class Room(Environment):
     """A single empty walled room - navix's `Empty` environments
     (`Navix-Empty-*`). One goal, one player, no obstacles; the task is
-    pure navigation. Reward and termination are the defaults: `+1` on
-    reaching the goal minus a per-step cost, episode ends on the goal.
+    pure navigation. `+1` on reaching the goal, `0` otherwise; the
+    episode ends on the goal.
 
     Attributes:
         random_start: `False` (the `Navix-Empty-NxN-v0` ids) places the
--- a/navix/environments/multi_room.py
+++ b/navix/environments/multi_room.py
@@ -535,7 +535,10 @@ def register_multi_room(env_id: str, num_rooms: int, max_room_size: int) -> None
             max_room_size=max_room_size,
             max_steps=kwargs.pop("max_steps", num_rooms * 20),
             observation_fn=kwargs.pop("observation_fn", observations.symbolic),
-            reward_fn=kwargs.pop("reward_fn", rewards.DEFAULT_TASK),
+            reward_fn=kwargs.pop(
+                "reward_fn",
+                rewards.compose(rewards.on_goal_reached, rewards.time_cost),
+            ),
             termination_fn=kwargs.pop("termination_fn", terminations.DEFAULT_TERMINATION),
             *args,
             **kwargs,
--- a/navix/environments/multi_room.py
+++ b/navix/environments/multi_room.py
@@ -113,6 +113,14 @@ MAX_LAYOUT_RESTARTS = 32  # a single room's own retry (MAX_PLACEMENT_TRIES,
 # restarts the *entire* room chain from scratch with a fresh key,
 # rather than trying to locally patch just the one room that failed.
 
+# MultiRoom keeps `DEFAULT_TASK`'s old shaped reward (goal `+1` minus a
+# per-step cost) even after `rewards.DEFAULT_TASK` dropped the cost.
+# Module-level (one shared object), so two MultiRoom instances of the
+# same id compare equal - `test_registry` checks a v1-alias redirect
+# builds an env `==` the direct one, which a fresh per-call `compose`
+# lambda would break.
+MULTI_ROOM_REWARD_FN = rewards.compose(rewards.on_goal_reached, rewards.time_cost)
+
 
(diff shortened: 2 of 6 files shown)


epignatelli:
@claude review

epignatelli:

The review ran against `b8609f9` (08:28) — the current head `2e6de64` (pushed 08:31) already handles its points:

- **MultiRoom (§1 ❌ / §2 / §3):** fixed before the review landed. `multi_room.py` now pins `reward_fn=compose(on_goal_reached, time_cost)`, so `Navix-MultiRoom-{N2-S4,N4-S5,N6}-v0` keep their `+1` / `-0.01`-per-step reward unchanged. New test `test_multiroom_keeps_its_per_step_cost` asserts a non-goal step still returns `-0.01`.
- **§4 breaking-changes footer:** now accurate — with MultiRoom pinned, no `Navix-*` id changes; the `BREAKING CHANGE:` footer covers only the `Environment.create` fallback and the `action_cost` deprecation. PR description updated to match.
- **§5 audit:** done — I instantiated all 83 registered envs and checked `env.reward_fn`; `multi_room.py` was the *only* `kwargs.pop("reward_fn", rewards.DEFAULT_TASK)` site. Every other env passes its own.
- **§5 `events.py` stale comment:** fixed in `2e6de64` — the `DONE_ACTION` comment no longer cites `action_cost` as precedent for hard-coding an index.

Re-review of `2e6de64` welcome.

epignatelli:

CI `Test (4)` failure fixed in `8312615`.

**Cause:** the previous commit built `compose(on_goal_reached, time_cost)` *inside* `register_multi_room`'s lambda, so each `MultiRoom.create` got a fresh `compose` closure. `test_registry.test_v1_aliases_redirect_and_warn` asserts `nx.make("MiniGrid-MultiRoom-N4-S5-v1") == nx.make("Navix-MultiRoom-N4-S5-v0")`, and two distinct lambda objects made the frozen `Environment` structs compare unequal (`reward_fn` field).

**Fix:** module-level `MULTI_ROOM_REWARD_FN` singleton (as `rewards.DEFAULT_TASK` was), referenced by all three registrations.

Locally: `test_registry` / `test_tasks` / `test_environments` / `test_events` (25) pass; `test_multi_room` (which the earlier bg run covered) passes.

PR #220, opened 2026-09-08
refactor(spaces): back Discrete/Continuous with distrax distributions

### LINKED ISSUE #219: Consider replacing navix/spaces.py with distrax distributions
## Summary

`navix/spaces.py` is a small bespoke descriptor hierarchy (`Space`,
`Discrete`, `Continuous`) with a hand-rolled `sample(key)`. `distrax` is
**already a hard dependency** (`requirements.txt`, used throughout
`agents/`, `experiment.py`, `es.py`, `benchmarks/`). We should evaluate
replacing the home-grown spaces with `distrax` distributions, or at least
delegating their sampling to `distrax`, to drop reimplemented code and get
`log_prob`/`entropy`/well-tested samplers for free.

Motivation trigger: #210 was a correctness bug in our own
`Discrete.sample` (exclusive-vs-inclusive `maxval`) - exactly the kind of
thing a vetted library sampler avoids.

## The catch: space descriptor vs. probability distribution

`distrax` models *probability distributions*, not *space descriptors*. Our
`Space` carries `shape` + `dtype` + element-wise `[minimum, maximum]` and
is used for introspection well beyond sampling:

- `Environment.observation_space / action_space / reward_space` - frozen
  `struct.field(pytree_node=False)` fields, compared by `==` (the
  `test_registry` v1-alias check relies on this equality).
- `Environment._get_obs_space_from_fn` infers a `Discrete`/`Continuous`
  from each `observation_fn` (`environments/environment.py:492-532`).
- `environments/wrappers.py` converts `observation_space` -> gymnax `Box`
  and `action_space` -> gymnax `Discrete` for the purejaxrl bridge.
- `Discrete.n` / `Discrete.create(len(action_set))`.
- Agents only actually call `env.observation_space.sample(rng)` to shape
  network init (`agents/ppo.py:357`, `agents/pqn.py:408`, dreamer, and the
  mirrored test files).

`distrax.Categorical` needs logits/probs and has no natural
`shape+dtype+bounds` contract for e.g. a `(H, W, 3) uint8 in [0, 255]`
image observation, so a naive 1:1 swap loses the descriptor role.

## Options to weigh

1. **Full replace.** `spaces` becomes thin factories returning `distrax`
   distributions (`Discrete` -> a discrete uniform `Categorical`;
   `Continuous` -> `distrax.Uniform`). Requires re-homing shape/dtype/bounds
   metadata (subclass the distrax dist, or carry a sidecar descriptor) and
   revisiting the frozen-pytree `==` semantics of distrax dists as
   `Environment` fields.
2. **Keep `Space` as the descriptor, delegate `.sample()` to distrax.**
   Minimal surface change: `Discrete.sample` -> `distrax.Categorical(
   logits=zeros(n)).sample(...)`, `Continuous.sample` -> `distrax.Uniform`.
   Keeps `==`, wrappers, `_get_obs_space_from_fn`, `.n` untouched. Gets the
   vetted sampler; skips `log_prob`/`entropy`.
3. **Do nothing** - `spaces.py` is ~180 lines and now correct post-#210;
   the descriptor need is real and distrax doesn't cover it.

## Affected surface (for whoever picks this up)

- `navix/spaces.py`, `tests/test_spaces.py`
- `navix/environments/environment.py` (fields, `create`, `_get_obs_space_from_fn`)
- `navix/environments/wrappers.py` (gymnax conversion)
- `navix/agents/{ppo,pqn,dreamer}.py` + `tests/test_{ppo,pqn}.py` (`.sample` for init)
- docs stubs render `::: navix.spaces`

## Recommendation

Lean towards **Option 2** unless we want `log_prob`/`entropy` off spaces
(we don't today - agents build their own `distrax.Categorical` from
network logits). Option 1 is a larger change whose main payoff is already
covered elsewhere. Decide before writing code.

_Filed from the #167 docs-pass follow-up thread._

epignatelli:
Closes #219 (the "full replace" option).

## What changed

`navix/spaces.py` no longer defines a bespoke `Space(struct.PyTreeNode)`
hierarchy with a hand-rolled `sample`. Instead:

- `Discrete` **subclasses `distrax.Categorical`** (uniform `1-D` logits of
  length `n_elements`).
- `Continuous` **subclasses `distrax.Uniform`**.
- `spaces.Space` is now an alias for `distrax.Distribution` — the common
  supertype the `Environment` type hints use.

So `isinstance(env.action_space, distrax.Distribution)` holds and
`log_prob` / `entropy` / `prob` / `mode` are available on any space.

## What stayed the same (so nothing downstream breaks)

Every attribute navix code reads off a space is preserved as a thin
override:

| surface | kept as |
|---|---|
| `space.shape` | navix attribute set in `__init__` |
| `space.dtype` | overridden property → the descriptor dtype |
| `space.minimum` / `space.maximum` | `0` / `n-1` for `Discrete`; `low` / `high` for `Continuous` |
| `space.n` (`Discrete`) | `num_categories` |
| `space.sample(key)` | positional key, returns a `space.shape`-shaped iid array |
| `space.replace(shape=…)` | rebuilds the space (used to flatten obs) |
| `Discrete.create` / `Continuous.create` | thin named constructors, unchanged call sites |
| `Space` symbol imported from `navix.spaces` | now `distrax.Distribution` |

`__eq__` / `__hash__` are defined by value on both classes, so
`nx.make(id) == nx.make(id)` still holds (the spaces are
`pytree_node=False` fields on the frozen `Environment` and land in the jit
treedef — `test_registry::test_v1_aliases_redirect_and_warn` covers this).

## Deliberate deviations from vanilla distrax

Trying the literal "spaces *are* distributions" turned up three places
where a faithful mapping is either wrong or wasteful, so the subclasses
deviate on purpose (all documented in the module docstring):

1. **Logits are `1-D` `(n_elements,)`, not `(*shape, n_elements)`.** A
   faithful event-shaped categorical for a `(56, 56, 3)` pixel
   observation would allocate a multi-MB zero-logits array *per
   environment* just to describe the space. The `1-D` distribution is
   "one element's distribution", and navix's array shape lives in
   `space.shape`. Consequence: `space.event_shape` / `space.batch_shape`
   are `()`, not `space.shape`.
2. **`sample(key)` keeps navix's positional-key signature**, not
   distrax's keyword-only `sample(*, seed, sample_shape)`. Call sites all
   pass the key positionally.
3. **`dtype` is overridden** to return the descriptor dtype (`distrax`
   otherwise infers it by `eval_shape`-ing `sample`, which our override
   defeats).

## Behaviour

**No environment behaviour change.** `Discrete.sample` /
`Continuous.sample` issue the same `jax.random.randint` / `.uniform` call
as before — same draws for the same key. This is a `refactor`, not a
`fix`/`feat`.

One narrow edge: the old `Discrete` tolerated `n_elements ~ 1e8` (it never
materialised anything); the new one allocates a length-`n_elements`
logits vector, so pathological counts cost memory. Real `n_elements`
(action counts, entity-tag counts, `256`) are trivially small.
`tests/test_spaces.py` was updated to reflect this and to cover the
distrax API, value equality, `replace`, and jit-treedef stability.

## Tests

`pytest tests/test_spaces.py tests/test_registry.py tests/test_environments.py
tests/test_memory.py tests/test_tasks.py tests/test_ppo.py tests/test_pqn.py
tests/test_dreamer.py tests/test_benchmark.py tests/test_agents.py
tests/test_models.py tests/test_experiment.py tests/test_issues.py` — all
pass. `examples/ppo.py`, `examples/ppo_pomdp.py`, `examples/hparam_search.py`
run clean. Full suite run before opening.

### DIFF
--- a/navix/__init__.py
+++ b/navix/__init__.py
@@ -52,8 +52,9 @@ Layout:
 - `navix.grid` - array helpers for grid geometry (cropping, rotation,
   line of sight).
 - `navix.rendering` - sprite/tile rendering for RGB observations.
-- `navix.spaces` - `Space` descriptors for observations, actions and
-  rewards.
+- `navix.spaces` - `Space` descriptors (shape / dtype / bounds) for
+  observations, actions and rewards; each is a `distrax` distribution
+  (`Discrete` <- `distrax.Categorical`, `Continuous` <- `distrax.Uniform`).
 - `navix.agents` - reference JAX implementations of PPO, PQN and
   DreamerV3, plus `navix.experiment.Experiment` to run and log them.
 - `navix.benchmarks` - experimental protocols that pin an environment,
--- a/navix/environments/environment.py
+++ b/navix/environments/environment.py
@@ -179,7 +179,8 @@ class Environment(struct.PyTreeNode):
             once `t >= max_steps`. Default (via `create`) is
             `4 * height * width`.
         observation_space: `Space` describing `observation_fn`'s output
-            (shape, dtype, bounds).
+            (shape, dtype, bounds). A `distrax` distribution - see
+            `navix.spaces`.
         action_space: `Discrete` over `len(action_set)`.
         reward_space: `Continuous` bound on the reward, `[-1, 1]` by
             default.
--- a/navix/spaces.py
+++ b/navix/spaces.py
@@ -14,9 +14,29 @@
 
 
 """`Space` descriptors for an environment's observation, action and
-reward arrays - shape, dtype and element-wise bounds, plus a `sample`
-that draws a conforming array. `Discrete` for integers, `Continuous` for
-floats.
+reward arrays.
+
+A space says three things about an array that flows through an
+environment - its `shape`, its `dtype`, and its element-wise
+`[minimum, maximum]` bounds - and can `sample` a conforming array.
+
+Each concrete space *is* a `distrax` distribution: `Discrete` subclasses
+`distrax.Categorical` (uniform over `0 .. n_elements - 1`) and
+`Continuous` subclasses `distrax.Uniform` (uniform over
+`[minimum, maximum]`). So a space also carries `log_prob`, `entropy`,
+`prob` and `mode` for free, and `isinstance(space, distrax.Distribution)`
+holds. `Space` re-exports `distrax.Distribution` as the common supertype
+navix annotates with.
+
+The distribution is over a *single* element: the underlying
+`distrax.Categorical.logits` is `1-D` (`(n_elements,)`), not
+`(*shape, n_elements)`, so the descriptor stays cheap even for a large
+image observation. `space.shape` / `space.dtype` are navix attributes on
+top of that; `space.event_shape` / `space.batch_shape` (the `distrax`
+introspection) are therefore `()`, not `space.shape`. `space.sample(key)`
+takes the key positionally (navix convention) and returns a
+`space.shape`-shaped array with iid elements - it does not follow
+`distrax`'s keyword-only `sample(*, seed, sample_shape)` signature.
 """
 
 from __future__ import annotations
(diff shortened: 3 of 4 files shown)


epignatelli:

Closing along with #219 (closed as not planned). This PR took the "subclass distrax" shape, which keeps a navix-specific space layer and had to override `batch_shape`/`event_shape`/`sample`/`dtype` to stay lightweight — not the direction wanted. See the #219 closing comment for the full rationale, including why the cleaner "delete `navix.spaces`, use plain distrax" version isn't worth doing now.

Issue #215, opened 2026-09-04
Remove leading-underscore names across `navix` (public-by-convention)


epignatelli:
_Filed by an AI agent on behalf of @epignatelli, from findings during the #167 documentation pass (see PR #209)._

Per the project convention that nothing in `navix` should be named with a leading underscore (raised again in the #167 thread), the following are still private and want renaming/exposing:

- `actions.py`: `_rotate`, `_can_walk_there`, `_move`
- `transitions.py`: `_can_spawn_there`
- `environments/crossings.py`: `_gaps_from_monotone_path`, `_crossings_obstacle_mask`, `_crossings_grid`
- `environments/key_corridor.py`: `_room_id`
- `environments/memory.py`: `_register_memory`
- `environments/obstructed_maze.py`: `_register_1d`, `_register_full`
- `environments/registry.py`: `_ENVS_REGISTRY`
- `environments/environment.py`: `_get_obs_space_from_fn`, and the `_reset` / `_step` override hooks on `Environment` + every environment subclass (this last one is the largest — decide whether the template-method hooks are in scope, since `reset` / `step` are the public wrappers).

Docstrings for these were intentionally left out of the #167 pass because `mkdocstrings` hides them; they should be documented as part of the rename.

PR #222, opened 2026-09-08
fix(observations): use MiniGrid's process_vis for first-person occlusion

### LINKED ISSUE #221: First-person observations use view_cone, not MiniGrid's process_vis, and diverge on FourRooms and SimpleCrossing
_Posted by an AI agent (Claude Opus 5, via Claude Code) on behalf of @xweinp._

## Summary

First-person observations mask with `view_cone`, which is a reachability flood rather than MiniGrid's visibility rule, so `categorical_first_person` and `rgb_first_person` do not reproduce MiniGrid's occlusion. On some shipped environments this changes what the agent observes.

## Where they disagree

Measured on `navix` 4.0.0 (`main` @ c1b3095), comparing the current `view_cone` mask against MiniGrid 3.1.0's `Grid.process_vis`, over **every standable agent pose × 4 directions** in six layouts per environment, restricted to the cells inside the first-person crop:

| environment | poses | cells in view | differing |
|---|---|---|---|
| `Navix-FourRooms-v0` | 6240 | 305760 | **3.44%** |
| `Navix-SimpleCrossingS11N5-v0` | 1128 | 55272 | **2.76%** |
| `Navix-Empty-16x16-v0` | 4704 | 230496 | 0.00% |
| `Navix-DoorKey-16x16-v0` | 4368 | 214032 | 0.00% |
| `Navix-KeyCorridorS4R3-v0` | 960 | 47040 | 0.00% |
| `Navix-KeyCorridorS6R3-v0` | 3648 | 178752 | 0.00% |

It goes in both directions: over 3000 random windows, 9165 cells were seen only by `view_cone` and 1501 only by `process_vis`.

## Minimal repro

`Navix-FourRooms-v0`, seed 0, agent at `(row=8, col=6)` facing `1` — 23 of the 49 cells differ. Windows are the `(2 * RADIUS + 1)` crop with the agent at the bottom-centre:

```
transparency            view_cone (current)     process_vis (MiniGrid)
    #......                 -------                 VVVVVVV
    #......                 -------                 VVVVVVV
    .......                 ---VVVV                 VVVVVVV
    #......                 ---VVVV                 VVVVVVV
    #......                 ---VVVV                 VVVVVVV
    ######.                 VVVVVVV                 VVVVVVV
    #......                 VVVVVVV                 VVVVVVV
```

The wall row has a single gap at the far right. MiniGrid propagates sideways without limit within a row, so sight flows through that gap and lights everything past it. `view_cone`'s diffusion does not reach the same cells.

To be clear about which is "right": MiniGrid's rule is itself very permissive — it is not line of sight, and it rounds corners. The issue is not that `view_cone` leaks; it is that NAVIX is a MiniGrid reimplementation whose observations get compared against MiniGrid's, and the two use different rules.

## Why it matters

`view_cone` also needs `RADIUS * 2` to cover the crop (see the comments at `observations.py:160` and `:348`), because it works in grid coordinates while the window is defined by the crop. Computing visibility in the cropped frame — where MiniGrid defines it — removes that coupling.

## Related, not covered here

`symbolic_first_person` applies no visibility mask at all, though its docstring says "Out-of-view / occluded cells are filled with the wall symbol". Filing separately per the scope rule in `AGENTS.md`.

## Proposed fix

PR to follow: add `grid.process_vis` (MiniGrid's rule, verified identical on 2880 + 500 random windows) plus `grid.first_person_view` wrapping it as a full-grid mask, and use it in both first-person observations.

xweinp:
_Posted by an AI agent (Claude Opus 5, via Claude Code) on behalf of @xweinp._

Fixes #221.

## What this changes

`categorical_first_person` and `rgb_first_person` masked with `view_cone`, a reachability flood through the full 8-neighbourhood. This replaces it with MiniGrid's own rule.

- **`grid.process_vis`** — MiniGrid's `Grid.process_vis` over an already-cropped first-person window. Visibility propagates one row at a time towards the front: sideways through transparent cells within a row (two passes, left to right then right to left), with every carrying cell also lighting the cell ahead and its two diagonal neighbours. MiniGrid's sequential inner loop is written here as two cumulative maxima per row, so a row is `O(log n)` parallel work rather than an unrolled `n`-step loop.
- **`grid.first_person_view`** — crops the transparency map, runs `process_vis` in the cropped frame (where MiniGrid defines visibility), and scatters the answer back to grid coordinates, so it is a drop-in for `view_cone` in the pipeline.
- Both observations now call it. It needs no doubled radius: the window *is* the crop, so the far row is covered by construction, which removes the `RADIUS * 2` coupling the old comments at `observations.py:160`/`:348` describe.

`view_cone` itself is untouched and still exported.

## ⚠️ This changes existing observations

Per `AGENTS.md`'s "Flag behavior-changing PRs explicitly": `categorical_first_person` and `rgb_first_person` return different values under unchanged environment IDs, wherever the two rules disagree. Measured over every standable agent pose × 4 directions in six layouts per environment:

| environment | differing cells in view |
|---|---|
| `Navix-FourRooms-v0` | 3.44% |
| `Navix-SimpleCrossingS11N5-v0` | 2.76% |
| `Navix-Empty-16x16-v0` | 0.00% |
| `Navix-DoorKey-16x16-v0` | 0.00% |
| `Navix-KeyCorridorS4R3-v0` | 0.00% |
| `Navix-KeyCorridorS6R3-v0` | 0.00% |

The second commit carries a `BREAKING CHANGE:` footer accordingly.

## Tests

New `tests/test_process_vis.py`, 10 tests. Every expectation in it was **generated and then re-verified by running real MiniGrid 3.1.0**, not written by hand:

- 8 golden windows pasted verbatim from `Grid.process_vis` output, plus 4 hand-built windows (gap in a wall, solid wall, empty room, boxed in) each re-derived with MiniGrid and confirmed to match.
- A further **500 independently drawn windows** were generated with MiniGrid and cross-checked against `grid.process_vis`: 0 mismatches. (Not committed — the suite pins behaviour with the pasted cases so it does not need MiniGrid as a test dependency.)
- Plus dtype (`bool` out, per `AGENTS.md`'s dtype discipline), `jit`/`vmap` equivalence, the round-trip between `first_person_view` and `process_vis`, and that sight stays inside the window and does not wrap around the array edges.

`tests/test_process_vis.py`, `test_observations.py`, `test_grid.py`, `test_entities.py` and `test_issues.py` pass. I did not run the full suite to completion — the PPO/PQN/Dreamer files exceed the memory available on this machine.

## Note on the repro in #221

MiniGrid's rule is itself permissive: it is not line of sight and it does round corners. The point of this PR is fidelity to MiniGrid, not stricter occlusion.


epignatelli:
Hey @xweinp ! That's a great catch, thank you for filing the issue. I just had a look at the review and left a couple of comments for your bot.

## 1. Correctness

- ✅ **`process_vis` reproduces MiniGrid's `Grid.process_vis`.** Re-derived independently against real MiniGrid 3.1.0: **0 cell mismatches over all 1,220 standable agent poses × 4 directions** in `Navix-FourRooms-v0` + `Navix-SimpleCrossingS11N5-v0` (the layouts that actually diverge). The PR's own 2,880 + 500 window checks agree.
- ✅ **Fixes #221.** With the new rule, `categorical_first_person` / `rgb_first_person` match a MiniGrid `process_vis` reference exactly on the poses that diverged on `main` — 0 disagreeing cells / RGB tiles on re-run.
- ✅ **"No doubled radius needed."** Correct: visibility is computed in the crop frame, so the far row is covered by construction and the old `RADIUS * 2` coupling is genuinely gone.
- ✅ **Intra-row `O(log n)` sweep.** The two-`cummax` formulation of MiniGrid's sequential inner loop is correct (confirmed by the end-to-end match above). The *outer* seven-row loop is still a `jax.lax.scan` — see §5.

## 2. Requested changes

- **`_sweep` breaks the module convention.** `navix/grid.py` has no other leading-underscore module-level name — every helper there (`crop`, `align`, `room`, `view_cone`, `draw_grid_lines`, …) is public — and the project's stated direction is to drop `_`-prefixed names entirely (#215). `tests/test_process_vis.py` reintroduces the same with `_parse`. Rename `_sweep` → `sweep`, or inline it into `process_vis`.
- **`first_person_view` is a crop → compute → scatter-back-to-`(H, W)` round trip, and the caller then crops again.** It exists only to leave `observations.py`'s world-frame masking untouched ("so the pipeline below is unchanged"). But `process_vis`'s own docstring says cropping the result afterwards is equivalent to running on the window — so apply `process_vis` to the *already-cropped* transparency map inside `categorical_first_person` / `rgb_first_person`, delete `first_person_view` and the second crop, and skip the full-grid intermediate. For `rgb_first_person` this also moves the mask *after* the crop, so the large `jnp.where` over the full `(H, W, TILE, TILE, 3)` patchwork shrinks to the `(7, 7, …)` window. Net: `grid.py` keeps one real primitive (`process_vis`), and the observation-pipeline plumbing stays in `observations.py` where it belongs.
- **Test provenance mismatch.** `tests/test_process_vis.py`'s docstring says the golden windows came from MiniGrid `v3.0.0`; the PR body and `feat(grid)` commit say `3.1.0`. Make it consistent.

## 3. New risks

- **First-person POMDP results shift for gapped layouts** under unchanged env IDs. Measured: `FourRooms` 3.4 % of observed cells, `SimpleCrossing` 2.8 %, `Empty` / `DoorKey` / `KeyCorridor` 0 %. Other interior-wall-with-gaps layouts (MultiRoom, LavaCrossing, ObstructedMaze, …) very likely shift too, though only those two were measured. Any recorded baseline numbers on these + `categorical_first_person` / `rgb_first_person` need refreshing.
- **Small-batch GPU slowdown.** As written (`lax.scan` over rows), a 64-step rollout at 128 envs is ~2× slower than `view_cone` on an A100 (categorical 14 → 31 ms, rgb 23 → 54 ms). It reaches parity by ~16 k parallel lanes (§5), so training-scale runs are fine, but small eval / debug loops on GPU regress. The §5 unroll removes this.
- **`process_vis(window)` hard-codes agent-at-bottom-centre** (`positions == cols // 2` + bottom-up scan) with no assertion — coupled to `crop`'s geometry by convention only. A future change to the crop layout would break it silently.

## 4. Breaking changes

- **`categorical_first_person` / `rgb_first_person` return different observations under the same env IDs**, wherever `view_cone` and `process_vis` disagree (numbers above). Correctly called out in the PR body, and the `fix(observations)` commit carries the `BREAKING CHANGE:` footer, so the CD major bump will be right.
- `navix.observations` no longer re-exports `view_cone` into its namespace (dropped from the `from .grid import …` list). `navix.grid.view_cone` is unchanged; only a caller doing `navix.observations.view_cone` would break. Negligible.
- `navix/_version.py` appears in the diff because the branch is based on the `4.0.0` release commit, not current `main`. CD rewrites `_version.py` on merge (per AGENTS.md), so it's noise — a rebase onto `main` would keep the diff to the three real files.

## 5. Broader opportunities

- **Unroll the row `lax.scan` into a plain Python `for`** — it is `2*RADIUS+1` = a static 7 iterations, so XLA compiles it straight-line. Benchmarked all three on an A100 (jax 0.11.0), 8 seeds × N envs, 32-step scan rollout, `categorical_first_person`; the loop's output is **bit-identical** to the scan (0 mismatches over every pose):

  | parallel lanes | `view_cone` | `process_vis` scan | `process_vis` py-loop | scratch scan → loop |
  |--:|--:|--:|--:|--:|
  | 1,024 | 17.3 ms | 17.6 ms | **9.2 ms** | 16 → 11 MB |
  | 16,384 | 56.4 ms | 55.4 ms | **51.2 ms** | 231 → 146 MB |
  | 65,536 | 212 ms | 228 ms | **150 ms** | 930 → 581 MB |
  | 131,072 | 439 ms | 440 ms | **285 ms** | 1860 → 1162 MB |

  The Python loop is **1.5–1.9× faster than both** the scan and `view_cone` at every scale, ~35 % less scratch memory, identical results, and it removes the §3 small-batch GPU regression. (On CPU the scan version is already 3–7× faster than `view_cone`; the loop keeps that.) The `feat(grid)` commit frames the `cummax` trick as avoiding "an unrolled Python loop" — true for the *inner* sweep, but the *outer* seven-row loop is exactly where the unroll wins.

- **`view_cone` becomes dead code in the library** — nothing under `navix/` references it after this PR (only `tests/test_grid.py` still tests it directly). Either keep it as a documented standalone utility or deprecate it. Relatedly, `tests/test_issues.py::test_147`'s comment still describes the old `view_cone` radius bug though the test now exercises `process_vis` via the obs fn.

- **`symbolic_first_person` still applies no visibility mask at all**, while its docstring claims occluded cells are filled with the wall symbol. After this PR it is the lone first-person observation not using `process_vis` — same root cause (noted out-of-scope in #221), worth tracking.

xweinp:

Thanks for the review — all three requested changes are in, plus the §5 unroll and the §3 assertion. Pushed as five commits on top of the branch rather than a rewrite, so the review's diff still reads.

**§2 Requested changes**

- **`_sweep` → `sweep`**, and `_parse` → `parse` in the tests (`refactor(grid): make sweep a public helper`).
- **`first_person_view` is gone** (`refactor(observations): compute first-person visibility on the crop`). Both observations now crop the transparency map and run `process_vis` on that window; no full-grid intermediate, no second crop. For `rgb_first_person` the mask moved after the crop, so the `jnp.where` runs over the `(7, 7, TILE, TILE, 3)` window instead of the full patchwork.
- **Test provenance**: the docstring now says 3.1.0, matching the PR body and the `feat(grid)` commit.

**One behaviour change that falls out of the crop-first rewrite**, called out because it is observable under unchanged env IDs and is *not* the occlusion fix:

Masking after the crop also blanks the off-map cells `crop` pads in. Those used to survive as the padding value `100`, which is not an `EntityId` and sits outside the `Discrete(MAX_CATEGORICAL_VALUE)` = `Discrete(11)` space `categorical_first_person` declares — that is true on `main` today, not something this branch introduced. They are now `UNKNOWN` (`0`), the tag `EntityIds` already documents for a cell a first-person view has not seen. Measured over every standable pose × 4 directions on FourRooms, SimpleCrossingS11N5, Empty-8x8, DoorKey-8x8 and KeyCorridorS4R3: `rgb_first_person` is bit-identical to the scatter-back version, and `categorical_first_person` differs **only** at those padding cells, always `100 → 0`. Pinned by `test_categorical_first_person_stays_inside_its_space`, and carried as a second `BREAKING CHANGE:` footer. Happy to split it out or revert to `100` if you would rather keep that separate from the occlusion fix.

**§3 — the un-asserted geometry**

`process_vis` now raises on an even-width window (`fix(grid): reject windows process_vis cannot place the agent in`), with the odd-width, agent-at-bottom-centre convention stated in a `Raises:` section. That is the case where there is no centre column at all, so a changed crop layout fails loudly instead of computing visibility from the wrong cell.

**§5 — took the unroll**

`perf(grid): unroll process_vis' row loop`. Confirmed bit-identical here too, end to end: the observation sweep above was run with the unrolled version and shows no difference beyond the padding cells. I have no A100 to reproduce your timings on, so the commit message quotes yours.

Also fixed `test_issues.py::test_147`'s comment, which still described the pure `view_cone` radius bug.

**Left alone, deliberately** — say the word if you want either in this PR:

- **`view_cone`** is still there. Deprecating it or keeping it as a documented standalone reads like your call, not mine.
- **`symbolic_first_person`** still applies no mask. Same root cause, but out of scope here and tracked in #221.

**Verification**: `process_vis` matches MiniGrid 3.1.0's `Grid.process_vis` on 4,000 freshly drawn random 7×7 windows, 0 mismatches, with the unroll and the rename in place. Local suite green — 24 test files, 130 tests, excluding the agent/benchmark files (`test_agents`, `test_benchmark`, `test_dreamer`, `test_ppo`, `test_pqn`, `test_models`, `test_experiment`), which train networks and are unrelated to this change.

epignatelli:

Re-review of the 5 follow-up commits. The structural points from the last round all landed cleanly — `sweep` is public, the row loop is a plain unrolled `for`, `first_person_view` is gone and visibility is computed on the crop, and there's a guard for the even-width case. Nice bonus: masking the crop instead of the full grid also fixes a real latent bug (crop's off-map padding `100` was leaking into `categorical_first_person`, outside its declared `Discrete` space), now `UNKNOWN(0)` with a test.

## 1. Correctness

- ✅ **`process_vis` still matches MiniGrid** after the `sweep`/unroll refactor — re-verified 0 mismatches over 1,260 poses (FourRooms + SimpleCrossing + MultiRoom-N2-S4), even-width guard raises, still `jit`s, the 28 touched tests pass.
- ✅ **crop-then-mask is equivalent** to the old mask-then-crop for seen cells, and correctly turns off-map padding into `UNKNOWN`. `test_categorical_first_person_stays_inside_its_space` pins it.
- ✅ Unroll is bit-identical to the previous `lax.scan` form (verified over every pose).

## 2. Requested changes

**The comments narrate history and comparisons instead of explaining the code.** Several new/edited blocks describe what the code *used to be* or *is better than* — `view_cone`, the deleted `lax.scan`, an old padding value — none of which a reader of the merged tree has in front of them. Cut the comparative clauses; keep the "why this code does this".

- `grid.py` `process_vis` docstring, middle paragraph ("This is deliberately not line of sight, and not `view_cone`. MiniGrid spreads one step sideways per row advanced, so it rounds corners a little; `view_cone` diffuses through the whole 8-neighbourhood every step, so it rounds them completely and reports cells behind a solid wall as seen. Matching MiniGrid exactly is the point…"). → one sentence: *"MiniGrid's rule verbatim — permissive, rounds corners, not line of sight."* The `view_cone` contrast belongs in the PR description.
- `grid.py`, the unroll comment — the last sentence *"Measured bit-identical to the scan, ~1.5-1.9x faster on GPU and a third less scratch memory."* is a benchmark against code that no longer exists. Drop it. Keep ~2 lines: *"Bottom-up (MiniGrid's order); each row needs the previous one. Static `2*RADIUS+1` trip count → plain loop, straight-line code XLA fuses across rows."*
- `observations.py` `categorical_first_person`, the view-mask comment — *"rather than view_cone's diffusion, which spreads through the whole 8-neighbourhood each step and so reports cells behind a solid wall as seen"* → drop. *"`process_vis` is defined on the egocentric window, so crop first and mask the crop. `padding_value=0` → off-grid reads as opaque."*
- `observations.py` `categorical_first_person`, the crop-then-mask comment — *"they used to survive as the padding value 100…"* → *"Mask after the crop so crop()'s off-map padding (100, not an EntityId) also becomes UNKNOWN(0)."*
- `grid.py` even-width guard comment — trim to two lines (agent stands bottom-centre by crop() convention; even width = a changed crop layout, not a window this answers for).

Smaller: `test_categorical_first_person_stays_inside_its_space`'s comment has the same tendency; and `rgb_first_person`'s `dark_cell_colour` comment (pre-existing, only re-wrapped here) is 10 lines — worth tightening while the diff is already touching it, but not this PR's job to fix.

## 3. New risks

- The crop-then-mask change makes `categorical_first_person` return `0` instead of `100` for off-map cells in view — a behaviour change independent of the occlusion-rule change, hitting any pose near a map edge. It's an improvement (the old value was outside the declared space) but it's a second axis of "same env ID, different observation" — see §4.

## 4. Breaking changes

- Unchanged from the last round: `categorical_first_person` / `rgb_first_person` return different observations under the same env IDs; the `fix(observations)` commit carries the `BREAKING CHANGE:` footer, so the CD bump is right. One wording gap: the footer says "wherever the two rules disagree", but the off-map `100 → 0` change (§3) also fires where the rules *agree* — broaden the footer text so it covers both.

## 5. Broader opportunities

- Unchanged: `view_cone` is now entirely unreferenced by `navix/` (only `tests/test_grid.py` still exercises it) — keep as a documented utility or deprecate. `symbolic_first_person` still applies no visibility mask while its docstring says it does. Neither is this PR's scope.

xweinp:

§2 is in, as one commit: `docs: explain the code rather than what it replaced` (b9a896a). Comment-only — the diff touches no executable line, and `ruff check` reports the same 7 pre-existing findings on the five files before and after.

Each of the five blocks you named is now the version you suggested, give or take wording:

- `process_vis` docstring: the `view_cone` paragraph is one sentence — "This is MiniGrid's rule verbatim: permissive, rounds corners, not line of sight."
- the unroll comment: three lines, "Bottom-up, MiniGrid's order: each row needs the one behind it. The trip count is a static `2 * RADIUS + 1`, so a plain loop unrolls into straight-line code XLA fuses across rows." The benchmark against the deleted scan is gone.
- `categorical_first_person`'s view-mask comment: "`process_vis` is defined on the egocentric window, so crop first and mask the crop. `padding_value=0` reads off-grid as opaque, so sight cannot leave the map and come back."
- its crop-then-mask comment: "Mask after the crop so `crop()`'s off-map padding (100, not an `EntityId`, outside the `Discrete(MAX_CATEGORICAL_VALUE)` space this observation declares) also becomes `UNKNOWN` (0)."
- the even-width guard: two lines, agent stands bottom-centre by `crop()`'s convention, an even width is a changed crop layout.

Plus the smaller ones, and two you did not flag, on the same principle: `rgb_first_person`'s view-mask comment (dropped its `view_cone` clause), the `dark_cell_colour` comment (10 lines to 5, keeping why the fill is opacity-adjusted and why a scalar suffices), `test_categorical_first_person_stays_inside_its_space`, `test_issues.py::test_147`, and three comments in `tests/test_process_vis.py` that referred to `view_cone` or to a roll-based flood.

**§4 — the footer wording.** Not done, and I would rather you chose. `d59f0e9`'s footer covers the occlusion change only, as you say; the off-map `100` to `0` change carries its own `BREAKING CHANGE:` footer on `482581e`, so both axes reach the release notes, just as two entries rather than one. Broadening `d59f0e9`'s text means rewriting that commit message and force-pushing the commits above it — no tree changes, and there are no inline review comments to orphan, but it does renumber the branch under your last two reviews. Say the word and I will do it; otherwise the second footer stands.

§5 unchanged: `view_cone` and `symbolic_first_person` left alone as your call and out of scope.

epignatelli:

`b9a896a` — the comment pass is done well: every narration/comparison clause is gone (`view_cone`, the deleted `lax.scan`, the benchmark line, "used to survive as 100"), and what's left explains why the code is shaped the way it is. Verified it's comments-only in `grid.py`/`observations.py` (no logic change); touched tests pass.

One thing to fix before merge, though:

- **`.claude/pytest.done` got committed** (contents: `EXIT=1`). It's a local Claude Code tooling artifact, not on `main`, added by this commit. Drop it from the PR. `.claude/` isn't in the repo's `.gitignore` either — worth adding a line while you're removing the file (a repo-wide gap this just surfaced, not strictly this PR's job).
- Separately, `EXIT=1` in that file suggests a local `pytest` run exited non-zero when it was written — probably stale/unrelated, but worth a glance to confirm nothing here actually fails locally.

Nothing else changed since the last review: correctness, the `BREAKING CHANGE:` footer, and the earlier §5 notes (`view_cone` now dead code in the library; `symbolic_first_person` still unmasked) all stand as before.

xweinp:

Both fixed in `3868b99`.

- **`.claude/pytest.done` is gone** from the tree, and `.gitignore` now has a `.claude/` entry (plus `.aider*`). Its sibling `.claude/pytest.log` had been caught by the existing `*.log` line, which is why only the marker file showed up — the directory itself had no entry, so the same class of file could land again. It is a new commit rather than an amend of `b9a896a`: that commit is already on the branch you reviewed, and rewriting it would have meant a force-push under your review.

- **`EXIT=1` was the missing optional test dependency, not a real failure.** `minigrid` is in `requirements_test.txt` but was not installed in the environment that wrote the marker, so `test_observations.py::test_rgb_first_person` and `test_issues.py::test_147` both died at `import gymnasium` during collection. With MiniGrid 3.1.0 installed, `tests/test_process_vis.py tests/test_observations.py tests/test_grid.py tests/test_issues.py tests/test_entities.py` is **31 passed, exit 0** — including those two, which are the ones that compare against a real MiniGrid render.

§4 (folding the off-map `100` to `0` note into `d59f0e9`'s `BREAKING CHANGE:` footer) still needs a force-push to reword that commit message, so it is still yours to call; `482581e`'s own footer covers that axis meanwhile.

xweinp:

Both points from the last review are fixed, in `3868b99`, and the §4 footer wording from the round before is in `de65da9`.

**`.claude/pytest.done`** is out of the tree, and `.gitignore` now carries `.claude/` (plus `.aider*`). Its sibling `.claude/pytest.log` had been caught by the existing `*.log` line, which is why only the marker file surfaced — the directory itself had no entry, so the same class of file could land again.

**The `EXIT=1` in that marker was the missing optional test dependency, not a real failure.** `minigrid` is in `requirements_test.txt` but was not installed in the environment that wrote it, so `test_observations.py::test_rgb_first_person` and `test_issues.py::test_147` — the two tests that compare against a real MiniGrid render — died on their in-body `import gymnasium`. With MiniGrid 3.1.0 installed, `tests/test_process_vis.py tests/test_observations.py tests/test_grid.py tests/test_issues.py tests/test_entities.py` is 31 passed, exit 0.

**§4, the breaking-change footer.** `de65da9` states the union of both counts in one `BREAKING CHANGE:`: the occlusion change (`process_vis` rather than `view_cone`, with the per-environment percentages), and, independently, `categorical_first_person` reporting off-map cells as `EntityIds.UNKNOWN` instead of the raw crop padding `100` — which fires for any pose with padding in view, whichever occlusion rule is in force. It is an empty commit carrying the footer rather than a reword of `d59f0e9`, so the commits you have already reviewed keep their hashes. Say the word if you would rather have it folded into `d59f0e9` and the branch force-pushed.

That leaves only the §5 items you scoped out: `view_cone` is now unreferenced inside `navix/`, and `symbolic_first_person` still applies no visibility mask.

PR #229, opened 2026-09-19
fix(actions): pickup is a no-op when the pocket is full

xweinp:
MiniGrid picks an object up only while `self.carrying is None`. navix overwrote the pocket, which left the held item at `DISCARD_PILE_COORDS` with nothing pointing at it: invisible in every observation, unreachable by `drop`, and no longer able to open the door it was collected for.

The pickup is now gated on an empty pocket, so a blocked attempt records no event, discards nothing and overwrites nothing.

Tests: one new case in `tests/test_actions.py`.


xweinp:
Folded into #228, which now carries both action fixes.

PR #230, opened 2026-09-19
fix(observations): report the pocket in categorical_first_person

xweinp:
Follow-up to #1 / epignatelli#227, which it is stacked on: `categorical_first_person` already masks with `process_vis`, but still reported `PLAYER` at the player's own cell, so nothing in the observation said the key had been picked up.

It now writes the carried entity's tag there, as MiniGrid's `gen_obs_grid` does. With an empty pocket the `PLAYER` tag stays: a free cell reads `0` in this encoding, the same value as "not seen", so writing MiniGrid's empty cell would hide the player's own position rather than mark it.

Tests: one new case in `tests/test_observations.py`.


xweinp:
Folded into #227, which now carries both observation fixes.

